PR #12, opened 2021-12-01
Day 1 solution

adpi2:


day1.scala
```diff
@@ -0,0 +1,26 @@
...
+  val input = Source.fromFile("input/day1").mkString
+  val answer = part1(input)
+  println(s"The solution is $answer")
+
+@main def part2(): Unit =
+  val input = Source.fromFile("input/day1").mkString
+  val answer = part2(input)
+  println(s"The solution is $answer")
+
+def part1(input: String): String =
```
mlachkar: should return a Int, I think it make more sense

day1.scala
```diff
@@ -0,0 +1,26 @@
...
+  val answer = part2(input)
+  println(s"The solution is $answer")
+
+def part1(input: String): String = 
+  val depths = input.split('\n').iterator.map(_.toInt)
+  val pairs = depths.sliding(2).map(arr => (arr(0), arr(1)))
+  pairs.count((prev, next) => prev < next).toString
+
+def part2(input: String): String =
+  val depths = input.split('\n').iterator.map(_.toInt)
```
vincenzobaz, line 23:
```suggestion
  val depths = input.linesIterator.map(_.toInt)
```

day1.scala
```diff
@@ -0,0 +1,26 @@
...
+  val answer = part1(input)
+  println(s"The solution is $answer")
+
+@main def part2(): Unit =
+  val input = Source.fromFile("input/day1").mkString
+  val answer = part2(input)
+  println(s"The solution is $answer")
+
+def part1(input: String): String = 
+  val depths = input.split('\n').iterator.map(_.toInt)
```
mlachkar, line 18: the input could be defirection a Seq[Int]

day1.scala
```diff
@@ -0,0 +1,26 @@
...
+  val pairs = depths.sliding(2).map(arr => (arr(0), arr(1)))
+  pairs.count((prev, next) => prev < next).toString
+
+def part2(input: String): String =
+  val depths = input.split('\n').iterator.map(_.toInt)
+  val sums = depths.sliding(3).map(arr => arr(0) + arr(1) + arr(2))
```
vincenzobaz, line 24:
```suggestion
  val sums = depths.sliding(3).map(_.sum)
```
if the list ha size three, we can use sum

tgodzik, line 24: Fun thing is that you don't actually need to sum everything, just compare element `i` with `i+3`

```suggestion
  val sums = depths.sliding(4).count(arr => arr(0) < arr(3))
```

day1.scala
```diff
@@ -0,0 +1,26 @@
...
+  val pairs = sums.sliding(2).map(arr => (arr(0), arr(1)))
```
mlachkar, line 25: you could call part1 on part2 after doing  val sums = depths.sliding(3).map(arr => arr(0) + arr(1) + arr(2)).
don't prefer to use sum method ?
depths.sliding(3).map(_.sum).

day1.scala
bishabosha, line 24: what about `.map(_.sum)`

day1.scala
adpi2: nice thanks

day1.scala
vincenzobaz:
```suggestion
  val depths = input.linesIterator.map(_.toInt)
```

day1.scala
bishabosha: @vincenzobaz found it first :)

adpi2: Oh that's crazy. It's a bit hard to understand but yes it is true.

I will explain that solution as a bonus in the website.

day1.scala
```diff
@@ -0,0 +1,26 @@
...
+  println(s"The solution is $answer")
+
+@main def part2(): Unit =
+  val input = Source.fromFile("input/day1").mkString
+  val answer = part2(input)
+  println(s"The solution is $answer")
+
+def part1(input: String): Int = 
+  val depths = input.split('\n').iterator.map(_.toInt)
+  val pairs = depths.sliding(2).map(arr => (arr(0), arr(1)))
```
vincenzobaz, line 19: THis could be equivalent to `depths zip depths.tail`

day1.scala
tgodzik: I wrote it down altogether as:
```
val prev = arr(i) + arr(i+1) + arr(i+2)
val next = arr(i+1) + arr(i+2) + arr(i+3)
```
which made me realize that the two middle products are just not needed

day1.scala
adpi2: oh yes, I don't think I will show it though

README.md
```diff
@@ -1,4 +1,45 @@
...
+- some explanation of our solutions
+- more solutions from the community
+
+## Setup
+
+We use Visual Studio Code with Metals to write Scala code, and scala-cli to compile and run it.
+
+You can follow these [steps](https://scalacenter.github.io/scala-advent-of-code/setup) to set up your environement.
+
+### How ot open in Visual Studio Code
```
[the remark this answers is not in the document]
adpi2: thanks

day1.scala
```diff
@@ -0,0 +1,26 @@
...
+  val answer = part2(input)
+  println(s"The solution is $answer")
+
+def part1(input: String): Int = 
+  val depths = input.linesIterator.map(_.toInt)
+  val pairs = depths.sliding(2).map(arr => (arr(0), arr(1)))
+  pairs.count((prev, next) => prev < next)
+
+def part2(input: String): Int =
+  val depths = input.linesIterator.map(_.toInt)
```
mlachkar, line 23: so don't you prefer to read and parse the file on the previous level, so this method cannot file (I mean function part1 and part2, would take seq[INT] and return an Int..

mlachkar: it's just a suggestion .. it wont change the code

day1.scala
mlachkar: a prefer sliding.. and it's an intersting method to talk about in the teaching part.

day1.scala
adpi2: yes it would be better

the reason I did otherwise is that the web solver (in the website) takes the raw input as parameter

day1.scala
julienrf: The benefits of `zip` is that the types guide you. It gives a collection of pairs that all have exactly two elements. You don’t have to use “unsafe” operations to access the elements of the pair.

day1.scala
vincenzobaz: This is the signature we imposed the execution in the browser

day1.scala
mlachkar: forgot about it!

PR #13, opened 2021-12-01
Add day 1

adpi2:
image


docs/puzzles/day1.md
```diff
@@ -0,0 +1,139 @@
...
+In part 1 we computed all pairs of consecutive elements.
+In part 2 we computed the sums of all consecutive three elements.
+
+In each case there is a notion of sliding window of size n, where n is 2 or 3.
+For example, the sliding window of size 3 of `Seq(10, 20, 30, 40, 50)` is:
+
+`Seq(Seq(10, 20, 30), Seq(20, 30, 40), Seq(30, 40, 50))`.
+
+
+Can we generalize this procedure in a method that compute a sliding window of some size n on any sequence of elements.
```
sjrd, line 72:
```suggestion
We can generalize this procedure in a method that computes a sliding window of some size n on any sequence of elements.
```

docs/puzzles/day1.md
```diff
@@ -0,0 +1,139 @@
...
+Such a method already exists in the Scala standard library under the name `sliding`. It returns an iterator of arrays.
+
+```
+$ Seq(10, 20, 30, 40, 50).sliding(5).toSeq
```
sjrd, line 76: Shouldn't this be
```suggestion
$ Seq(10, 20, 30, 40, 50).sliding(3).toSeq
```
?

docs/puzzles/day1.md
```diff
@@ -0,0 +1,139 @@
...
+The second step and hardest challenge of this puzzle is to compute all pairs of consecutive depth measurments.
+
+For each index `i` from 0 until `depths.size - 1` we can create a pair of the depths at index `i` and `i + 1`.
+```scala
+val pairs: Seq[(Int, Int)] =
+  for i <- 0 until depths.size -1
+  yield (depths(i), depths(i + 1))
+```
+
+Then we can count the pairs whose first element is smaller than its second element.
```
sjrd, line 29: Perhaps, before this sentence, show an example of input and what `pairs` is for that input. That should help people visualize what is happening.

sjrd (approved):

docs/puzzles/day1.md
mlachkar: yep!

mlachkar (approved):
Nice!

PR #19, opened 2021-12-02
Day2

mlachkar:


src/day2.scala
```diff
@@ -0,0 +1,63 @@
+// using scala 3.0.2
+
+package day2
+
+import scala.io.Source
+
+@main def part1(): Unit =
+  val input = Source.fromFile("input/day2").mkString
```
julienrf, line 8: This is something I missed yesterday, but good practice is to `close()` the source after you have consumed it. I suggest using `Using`, which does this automatically for you. It would be something like:

```suggestion
  val input = util.Using.resource(Source.fromFile("input/day2"))(_.mkString)
```

src/day2.scala
```diff
@@ -0,0 +1,63 @@
...
+  val answer = part1(input)
+  println(s"The solution is $answer")
+
+@main def part2(): Unit =
+  val input = Source.fromFile("input/day1").mkString
```
julienrf, line 13: Same comment here regarding resource management.

src/day2.scala
```diff
@@ -0,0 +1,63 @@
...
+case class Position(x: Int, y: Int):
+  def +(p: Direction): Position =
+    p match
+      case Direction.Forward(newX) => Position(x + newX, y)
+      case Direction.Down(newY)    => Position(x, y + newY)
+      case Direction.Up(newY)      => Position(x, y - newY)
+
+  def result = x * y
+
+enum Direction(x: Int, y: Int):
```
adpi2, line 52: It seems x and y are not used so you should be able to remove them.

adpi2: Also Direction is called Command in the puzzle description

src/day2.scala
```diff
@@ -0,0 +1,63 @@
...
+    p match
+      case Direction.Forward(newX) =>
+        PositionWithAim(x + newX, y + newX * aim, aim)
+      case Direction.Down(newY) => PositionWithAim(x, y, aim + newY)
+      case Direction.Up(newY)   => PositionWithAim(x, y, aim - newY)
+
+  def result = x * y
+
+case class Position(x: Int, y: Int):
+  def +(p: Direction): Position =
```
julienrf, line 44: Code style preference:

```suggestion
  def + (p: Direction): Position =
```

Should we use `infix`? I never remember…

src/day2.scala
```diff
@@ -0,0 +1,63 @@
...
+  def +(p: Direction): PositionWithAim =
+    p match
+      case Direction.Forward(newX) =>
+        PositionWithAim(x + newX, y + newX * aim, aim)
+      case Direction.Down(newY) => PositionWithAim(x, y, aim + newY)
+      case Direction.Up(newY)   => PositionWithAim(x, y, aim - newY)
+
+  def result = x * y
+
+case class Position(x: Int, y: Int):
```
adpi2, line 43: In the puzzle description x is called `horizontal` and y is called  `depth`

src/day2.scala
```diff
@@ -0,0 +1,63 @@
...
+def part1(input: String): Int =
+  val entries = input.linesIterator.flatMap(Direction.from)
+  val firstPosition = Position(0, 0)
+  val lastPosition = entries.foldLeft(firstPosition)((position, direction) =>
+    position + direction
+  )
+  lastPosition.result
+
+def part2(input: String): Int =
+  val entries = input.linesIterator.flatMap(Direction.from)
```
julienrf, line 26: What does the specification says? Are you supposed to silently ignore parsing failures, or stop the program?

I personally think it is bad practice to swallow parsing failures like so. I would rather make them stop the program early, instead of producing an empty list of `entries`.

src/day2.scala
```diff
@@ -0,0 +1,63 @@
...
+
+case class PositionWithAim(x: Int, y: Int, aim: Int):
+  def +(p: Direction): PositionWithAim =
+    p match
+      case Direction.Forward(newX) =>
+        PositionWithAim(x + newX, y + newX * aim, aim)
+      case Direction.Down(newY) => PositionWithAim(x, y, aim + newY)
+      case Direction.Up(newY)   => PositionWithAim(x, y, aim - newY)
+
+  def result = x * y
```
julienrf, line 41: Should `result` be named `area` or something similar?

julienrf (commented):
This is a very easy read, I love this solution! I have left a couple of remarks, notably regarding resource management, and error handling.

src/day2.scala
mlachkar: I don't like either.
there is no method in the standard library to do Seq[Option] to Option[Seq]
I m planning to change that ^^

julienrf: In this situation, I would suggest failing early, instead of dealing with `Option`.

src/day2.scala
mlachkar: It's not the name used on the game

src/day2.scala
```diff
@@ -0,0 +1,63 @@
...
+  val input = Source.fromFile("input/day1").mkString
+  val answer = part2(input)
+  println(s"The solution is $answer")
+
+def part1(input: String): Int =
+  val entries = input.linesIterator.flatMap(Command.from)
+  val firstPosition = Position(0, 0)
+  val lastPosition = entries.foldLeft(firstPosition)((position, direction) =>
+    position + direction
+  )
```
julienrf, line 22: I think an interesting discussion point could be to compare this with an eta-expanded iteration function:

```suggestion
  def move(position: Position, command: Command): Position = ...
  val lastPosition = entries.foldLeft(firstPosition)(move)
```

PR #20, opened 2021-12-02
Day2 puzzle

mlachkar:


docs/puzzles/day2.md
```diff
@@ -0,0 +1,201 @@
...
+by [@mlachkar](https://twitter.com/meriamLachkar)
+
+## Puzzle description
+
+https://adventofcode.com/2021/day/2
+
+## Solution of Part 1
+
+### Parsing the file
+The first step is to model the problem correctly and then parse the input file.
```
adpi2, line 15:
```suggestion
The first step is to model the problem and to parse the input file.
```

docs/puzzles/day2.md
```diff
@@ -0,0 +1,201 @@
...
+  def from(s: String): Command =
+    s match
+      case s"forward $x" if x.toIntOption.isDefined => Forward(x.toInt)
+      case s"up $x"      if x.toIntOption.isDefined => Up(x.toInt)
+      case s"down $x"    if x.toIntOption.isDefined => Down(x.toInt)
+      case _ => throw new Exception(s"value $s is not valid command")
+
+  def parseFile(): Iterator[Command] = 
+    val input = util.Using.resource(Source.fromFile("input/day2"))(_.mkString)
+    input.linesIterator.map(Command.from)
```
adpi2, line 35: Not sure it adds anything here

docs/puzzles/day2.md
```diff
@@ -0,0 +1,201 @@
...
+      case s"down $x"    if x.toIntOption.isDefined => Down(x.toInt)
+      case _ => throw new Exception(s"value $s is not valid command")
+
+```
+
+## Enum in Scala 3
+An enumeration is used to define a type consisting of a set of named values.
+
+We used to used `sealed trait` with `case class` and `case object` to model this set of named values.
+The syntax is more consise and easier to explain than the combination we had in Scala 2.
```
adpi2, line 145: Scala 3 enums are more concise and easier to read that the Scala 2 ADTs.

```
enum Command:
  case Forward(x: Int)
  case Down(x: Int)
  case Up(x: Int)
```

is almost the same as:
```
sealed trait Command
object Command {
  case class Forward(x: Int) extends Command
  case class Down(x: Int) extends Command
  case class Up(x: Int) extends Command
}
```

docs/puzzles/day2.md
```diff
@@ -0,0 +1,201 @@
...
+An enumeration is used to define a type consisting of a set of named values.
+
+We used to used `sealed trait` with `case class` and `case object` to model this set of named values.
+The syntax is more consise and easier to explain than the combination we had in Scala 2. 
+
+Read [the official documentation](https://docs.scala-lang.org/scala3/reference/enums/enums.html) 
+for more details.
+
+## FoldLeft 
+`foldLeft` is a method from the standard library on `IterableOnce` (ie: Seq, List, Iterator, ...).
```
adpi2, line 151:
```suggestion
`foldLeft` is a method from the standard library on iterable collections (ie: `Seq`, `List`, `Iterator`, ...).
```

adpi2 (changes requested):
Could you add the button `Day 2` on the front page, please

docs/puzzles/day2.md
```diff
@@ -0,0 +1,201 @@
...
+```scala
+// signature of foldLeft
+def foldLeft[B](initialElement: B)(op: (B, A) => B): B
+
+// List sum
+val nums = List(1, 3, 2, 4)
+// equivalent to (((0 + 1) + 3 ) + 2 ) + 4 
+nums.foldLeft(0)((acc, cur) => acc + cur) // 10
+
+```
```
adpi2, line 164:
```suggestion
The signature of foldLeft is:
`scala
def foldLeft[B](initialElement: B)(op: (B, A) => B): B
`

`scala
List(1, 3, 2, 4).foldLeft(0)((acc, cur) => acc + cur) // 10
`

is the same as:
`scala
(((0 + 1) + 3) + 2) + 4
`
```

adpi2 (approved):

PR #23, opened 2021-12-03
Solution of day 3

sjrd:


src/day3.scala
```diff
@@ -0,0 +1,85 @@
...
+    keepMostCommon: Boolean): BitLine =
+  bitLines match
+    case Nil =>
+      throw new AssertionError("this shouldn't have happened")
+    case onlyLine :: Nil =>
+      onlyLine
+    case _ =>
+      val (bitLinesWithOne, bitLinesWithZero) =
+        bitLines.partition(line => line(bitPosition) == 1)
+      val onesAreMostCommon = bitLinesWithOne.sizeCompare(bitLinesWithZero) >= 0
```
julienrf, line 79: FYI, there is also:
```suggestion
      val onesAreMostCommon = bitLinesWithOne.sizeIs >= bitLinesWithZero
```

src/day3.scala
```diff
@@ -0,0 +1,85 @@
...
+  oxygenGeneratorRating * co2ScrubberRating
+
+@scala.annotation.tailrec
+def recursiveFilter(bitLines: List[BitLine], bitPosition: Int,
+    keepMostCommon: Boolean): BitLine =
+  bitLines match
+    case Nil =>
+      throw new AssertionError("this shouldn't have happened")
+    case onlyLine :: Nil =>
+      onlyLine
```
julienrf, line 75:
```suggestion
    case lastLine :: Nil =>
      lastLine
```
(not important)

julienrf (approved):
Brilliant!

Just a couple of minor suggestions.

Also, I think the program would benefit from more type ascription. E.g., `val bitLines: List[BitLine] = ...`, `val sumsOfOneBits: BitLine = ...`, etc.

src/day3.scala
sjrd: That does not compile. `bitLinesWithZero` is another collection. To use `sizeIs >= ...` it would have to be an `Int`.

src/day3.scala
sjrd: I've used `lastRemainingLine`.

src/day3.scala
julienrf: Oops, en effet.

mlachkar (approved):

mlachkar:
Let's merge it, and create the day3 page

PR #26, opened 2021-12-04
Add my solutions for Day 2 and Day 3

s5bug:
All of my solutions for each day are in the same place, so I just copied what someone else wrote for my Day 1 solution.


adpi2 (approved):
Cool, thanks!

PR #31, opened 2021-12-06
reinstate unintentionally removed line break before 'Share your solut…

philipschwarz:
I think I need to put back that lonely dash that I removed.
image


adpi2 (approved):
Thanks for noticing. A simple line break will do the job.

PR #34, opened 2021-12-06
Solution for day 6 of advent of code

julienrf:


src/day6.scala
```diff
@@ -0,0 +1,81 @@
...
+object Fish:
+  // "Suppose you were given the following list:
+  //
+  // 3,4,3,1,2
+  //
+  // This list means that the first fish has an internal timer of 3, the second fish
+  // has an internal timer of 4, and so on until the fifth fish, which has an
+  // internal timer of 2."
+  def parseSeveral(input: String): Seq[Fish] =
+    for timerStr <- input.split(",")
```
sjrd, line 37: There is a warning on this line:
```
[warn] .\src/day6.scala:39:5: method copyArrayToImmutableIndexedSeq in class LowPriorityImplicits2 is deprecated since 2.13.0: Implicit conversions from Array to immutable.IndexedSeq are implemented by copying; Use the more efficient non-copying ArraySeq.unsafeWrapArray or an explicit toIndexedSeq call
[warn]     for timerStr <- input.split(",")
[warn]     ^
```
It should be addressed.

src/day6.scala
```diff
@@ -0,0 +1,81 @@
...
+import scala.io.Source
+
+@main def part1(): Unit =
+  println(s"The solution is ${part1(readInput())}")
+
+@main def part2(): Unit =
+  println(s"The solution is ${part2(readInput())}")
+
+def readInput(): String =
+  Using.resource(Source.fromFile("input/day6"))(_.mkString.trim)
```
sjrd, line 14: It would be best to do the `trim` inside `parseSeveral`. The Scala.js UI in the webpage directly calls `part1(input)` rather than `part1()`, so if a visitor appends a newline in the input box, parsing will fail.

src/day6.scala
```diff
@@ -0,0 +1,81 @@
...
+// Use a cache to memoize the number of fishes created by one fish for some
+// number of remaining days to live
+val cache = mutable.Map.empty[Int, BigInt]
+
+def descendants(remainingDays: Int): BigInt =
+  cache.getOrElseUpdate(
+    remainingDays,
+    if remainingDays <= 0 then 1
+    else descendants(remainingDays - 7) + descendants(remainingDays - 9)
+  )
```
sjrd, line 81: The recursive calls to `descendants` modify the `cache`, in the middle of the `getOrElseUpdate`. This basically amounts to *concurrent modifications*. While [`mutable.Map.getOrElseUpdate`](https://www.scala-lang.org/api/2.13.7/scala/collection/mutable/Map.html#getOrElseUpdate(key:K,op:=%3EV):V) says that concurrent modifications have a specified behavior, it is not good enough for those purposes:

> Concurrent map implementations may evaluate the expression op multiple times, **or may evaluate op without inserting the result**.

(emphasis mine)

I don't know if `getOrElseUpdate` can be safely used in this context. I think that, to be safe, you need to use `get`, then compute recursive calls, then `update`.

sjrd (commented):
Who doesn't like an opportunity to show off memoization? 😅

src/day6.scala
julienrf: How did you get the warning? Scala-cli does not complain.

adpi2: There is a known bug in Bloop that could explain why scala-cli does not complain. You should see it if you do `scala-cli clean .` then `scala-cli compile .`.

sjrd: It was scala-cli that complained for me. Latest scala-cli, on Windows.

julienrf: OK, funny, I did run `scala-cli clean src/day6.scala` but the warning did not show up. However, after running `scala-cli clean src/` then I can see the warning…

src/day6.scala
julienrf: That’s a good point. I didn’t check that.

That being said, I think the code I wrote stays correct. The spec says that it could happen that some results are not effectively cached. In our case, this does not have an impact on the correctness of the algorithm.

I tried moving the cache management outside of the recursive calls but then the performance was 10 times worse than the current implementation (143.7 seconds vs 1.3 s), so, I would just keep the current one.

src/day6.scala
julienrf: Thank you guys. This is fixed in 46e4344

src/day6.scala
julienrf: Fixed in 46e4344

julienrf:
I pushed https://github.com/scalacenter/scala-advent-of-code/pull/34/commits/117934bba2b6b082dbe2182d8db57682cbfc9228 with an alternative approach suggested by @sjrd that seems more elegant to me. What do you think?

src/day6.scala
```diff
@@ -0,0 +1,88 @@
...
+    .sum
+
+def tick(population: Map[Int, BigInt]): Map[Int, BigInt] =
+  def countPopulation(daysLeft: Int): BigInt = population.getOrElse(daysLeft, BigInt(0))
+  Map.from(
+    for day <- 0 until 9
+    yield
+      val newPopulation =
+        if day == 6 then countPopulation(7) + countPopulation(0)
+        else countPopulation((day + 1) % 9)
```
sjrd, line 86: I would suggest to use an additional branch for 8 instead of the trick with `%`:
```suggestion
      val newPopulation =
        if day == 6 then countPopulation(7) + countPopulation(0)
        else if day == 8 then countPopulation(0)
        else countPopulation(day + 1)
```
The fact that `day == 8` fetches `countPopulation(0)` is much more related to the fact that `day == 6` has to do so as well. Grouping those two as the special cases makes more sense to me than shoehorning the case 8 with the "regular" cases.

julienrf: Should I use a loop at all, then?

~~~ scala
Map(
  0 -> countPopulation(1),
  1 -> countPopulation(2),
  2 -> countPopulation(3),
  3 -> countPopulation(4),
  4 -> countPopulation(5),
  5 -> countPopulation(6),
  6 -> countPopulation(7) + countPopulation(0),
  7 -> countPopulation(8),
  8 -> countPopulation(0)
)
~~~

sjrd: Good point. It's even clearer like that.

mlachkar: or add a comment to explain how the trick deal with that

julienrf: Solved in a825ddf

PR #36, opened 2021-12-06
Add solution for day 5

tgodzik:


docs/puzzles/day5.md
```diff
@@ -2,12 +2,206 @@
...
+
+  map.count { case (_, v) => v > 1 }
+end findDangerousPoints
+```
+
+### Hydrothermal Venture 2: The Diagonal Menace
+
+For the second part of the puzzle we need to take into account also the diagnal
+vents. Fortunately, they can only have an angle of 45 degrees, which means when
+iterating over ranges both x and y positions increment by 1. So we can add
```
adpi2, line 156: I find this sentence hard to read:
```suggestion
vents. Fortunately, they can only have an angle of 45 degrees, which means that
both x and y positions increment by 1 at each step of the range. So we can add
```

docs/puzzles/day5.md
```diff
@@ -2,12 +2,206 @@
 sidebar_position: 5
 ---
 
+by [@tgodzik](https://github.com/tgodzik)
+
```
adpi2, line 6: Maybe you need to move that after the title `# Day 5: Hydrothermal Venture`.

docs/puzzles/day5.md
tgodzik: I was writing it a bit in a hurry :sweat:

docs/puzzles/day5.md
tgodzik: Done!

PR #37, opened 2021-12-07
Add solution for day 7

tgodzik:


src/day7.scala
```diff
@@ -0,0 +1,76 @@
...
+
+package day7
+
+import scala.util.Using
+import scala.io.Source
+import scala.annotation.tailrec
+
+extension (s: String)
+  def input: List[Int] =
+    Using.resource(Source.fromFile("input/day7"))(_.mkString).split(",").map(_.toInt).toList
```
adpi2, line 12: It seems that `s` is not used and so the extension is not really useful

tgodzik: ach damn, I was converting my solution on the fly based on what I saw in @julienrf 's one

src/day7.scala
```diff
@@ -0,0 +1,76 @@
...
+          if p.horizontal == minHorizontal.horizontal then p.moveUp()
+          else p
+        }
+        align(updated, fuelCost + fuelCostForMin)
+    end if
+  end align
+end Crabmada
+
+@main
+def main =
```
sjrd, line 63: In order to integrate with the website, you will need separate methods `def part1(input: String): Int` and `def part2(input: String): Int`.

src/day7.scala
tgodzik: Fixed!

src/day7.scala
```diff
@@ -0,0 +1,74 @@
...
+import scala.util.Using
+import scala.io.Source
+import scala.annotation.tailrec
+
+case class Crabmarine(
+    horizontal: Int,
+    increasingCost: Boolean,
+    fuelCost: Int = 1
+):
+  def moveUp() =
```
sjrd, line 15: Shouldn't this be called `moveForward` and the other one `moveBackward`?

src/day7.scala
```diff
@@ -0,0 +1,74 @@
...
+      val minHorizontal = situation.minBy(_.horizontal)
+
+      val fuelCostForTop = situation.collect {
+        case p if p.horizontal == topHorizontal.horizontal => p.fuelCost
+      }.sum
+      val fuelCostForMin = situation.collect {
+        case p if p.horizontal == minHorizontal.horizontal => p.fuelCost
+      }.sum
+      if fuelCostForTop < fuelCostForMin then
+        val updated = situation.map { p =>
```
julienrf, line 46: In general, I suggest using longer names. I am not sure what `p` stands for here. It seems to be a `Crabmarine`, what about `crabmarine`, or `crab`, instead?

src/day7.scala
```diff
@@ -0,0 +1,74 @@
...
+  @tailrec
+  final def align(
+      situation: List[Crabmarine] = crabmarines,
+      fuelCost: Int = 0
+  ): Int =
+    val allTheSame = situation.forall(_.horizontal == situation.head.horizontal)
+    if allTheSame then fuelCost
+    else
+      val topHorizontal = situation.maxBy(_.horizontal)
+      val minHorizontal = situation.minBy(_.horizontal)
```
sjrd, line 37: Perhaps `maxHorizontal` to go with `minHorizontal`?

src/day7.scala
```diff
@@ -0,0 +1,74 @@
...
+        align(updated, fuelCost + fuelCostForMin)
+    end if
+  end align
+end Crabmada
+
+@main
+def main =
+  val input = 
+    Using.resource(Source.fromFile("input/day7"))(_.mkString).split(",").map(_.toInt).toList
+  val crabmarines1 = input.map(p => Crabmarine(p, increasingCost = false))
```
julienrf, line 66: `p` again? This time it seems to be an `Int`, and it is used to fill the `horizontal` field of the case class. What about the following?

```suggestion
  val crabmarines1 = input.map(horizontal => Crabmarine(horizontal, increasingCost = false))
```

sjrd (commented):
I'm not sure I understand *why* this works, so I'm looking forward to the accompanying post :)

src/day7.scala
```diff
@@ -0,0 +1,74 @@
...
+
+package day7
+
+import scala.util.Using
+import scala.io.Source
+import scala.annotation.tailrec
+
+case class Crabmarine(
+    horizontal: Int,
+    increasingCost: Boolean,
```
julienrf, line 12: Design-wise, I wonder if instead of having a field `increasingCost` we should have two specialized classes with different implementations for `moveUp` and `moveDown`?

julienrf (commented):
Nice job Tomasz!

src/day7.scala
tgodzik: Thanks! Fixed!

src/day7.scala
tgodzik: Makes sense, since it's horizontal

src/day7.scala
tgodzik: Changed

src/day7.scala
tgodzik: Renamed!

src/day7.scala
tgodzik: Changed!

julienrf (approved):

src/day7.scala
tgodzik: Refactored it, looks much nicer :100:

tgodzik:
Apparently the solution can also be median for P1 and average for P2, but I think this might be harder to explain maybe ?

Or should I revise the solution?

julienrf:
As you feel 😛

src/day7.scala
```diff
@@ -0,0 +1,89 @@
...
+      situation: List[Crabmarine] = crabmarines,
+      fuelCost: Int = 0
+  ): Int =
+    val allTheSame = situation.forall(_.horizontal == situation.head.horizontal)
+    if allTheSame then fuelCost
+    else
+      val maxHorizontal = situation.maxBy(_.horizontal)
+      val minHorizontal = situation.minBy(_.horizontal)
+
+      val fuelCostForTop = situation.collect {
```
sjrd, line 51:
```suggestion
      val fuelCostForMax = situation.collect {
```
to be consistent everywhere.

sjrd (approved):

src/day7.scala
tgodzik: Fixed!

PR #39, opened 2021-12-08
Add solution of day 4 by @sporarum

adpi2:
@sporarum


src/day4.scala
```diff
@@ -0,0 +1,55 @@
+// using scala 3.0.2
+
+package day4
+
+import scala.io.Source
+
+@main def run(): Unit =
```
sjrd, line 7: You'll have to follow the format of `part1` and `part2` with 2 overloads each that the other solutions use. This is necessary to integrate in the website afterward.

src/day4.scala
```diff
@@ -0,0 +1,55 @@
...
+  
+  val boards: List[Board] = 
+    inputSections.tail.map(Board.parse).map(_.mapNumbers(numberToTurn))
+  
+  // for each board, the number of turns until it wins
+  val winningTurns: List[(Board, Int)] =
+    for board <- boards
+    yield
+      val lineMin = board.lines.map(line => line.max).min
+      val  colMin = board.columns.map(col => col.max).min
```
sjrd, line 41: That is a very unusual formatting choice. There are two common choices in Scala. Either without any alignment:
```suggestion
      val colMin = board.columns.map(col => col.max).min
```
or with alignment of `=`, but the rest stays aligned on the left, never on the right:
```suggestion
      val colMin  = board.columns.map(col => col.max).min
```

src/day4.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  private val lineParser = raw"\d+".r
+  private def parseLine(inputLine: String): List[Int] =
+    lineParser.findAllIn(inputLine).toList.map(_.toInt)
+
+def answers(input: String): (Int, Int) =
+  val inputSections: List[String] = input.split("\n\n").toList
+  val numbers = inputSections.head.split(',').map(_.toInt)
+
+  val numberToTurn = numbers.zipWithIndex.toMap
```
sjrd, line 30: This only works if numbers are never drawn twice. This is not guaranteed by the problem statement, so we should not rely on it. We must only keep the first turn at which any given number is drawn.

src/day4.scala
```diff
@@ -0,0 +1,55 @@
...
+  val turnToNumber = numberToTurn.map((key, value) => (value, key))
+  
+  val boards: List[Board] = 
+    inputSections.tail.map(Board.parse).map(_.mapNumbers(numberToTurn))
```
sjrd, line 34: This will throw an exception if a number of one of the boards is *never* drawn. I would expect many datasets to have that case happening, and am a bit surprised it worked with your dataset.

sjrd (commented):
Nice solution overall. :)

I have a few comments.

src/day4.scala
adpi2: I discussed it with @sporarum and we concluded that it was easier to solve both parts in a single function. 

For the website part we will have a single box and button that will print the solution of both parts.

src/day4.scala
adpi2: Indeed it is not stated in the puzzle description but we assume it is implicit. Indeed it seems that everyone's input contains all numbers from `0` to `99`. Even in the example the input contains all numbers from `0` to `26`. @sporarum will state it clearly in its blog post.

src/day4.scala
adpi2: Does it throw an exception with your dataset?

PR #41, opened 2021-12-08
complete solution for Day 8

bishabosha:


src/day8.scala
```diff
@@ -0,0 +1,130 @@
...
+  case Five extends Digit(A, B, D, F, G)
+  case Six extends Digit(A, B, D, E, F, G)
+  case Seven extends Digit(A, C, F)
+  case Eight extends Digit(A, B, C, D, E, F, G)
+  case Nine extends Digit(A, B, C, D, F, G)
+
+object Digit:
+
+  private val uniqueLookup =
+    values.groupBy(_.segments.size).collect { case k -> Array(d) => k -> d }
```
julienrf, line 125: It would be nice to write the type of `uniqueLookup`

src/day8.scala
```diff
@@ -0,0 +1,130 @@
...
+end substitutions
+
+enum Segment:
+  case A, B, C, D, E, F, G
+
+  val char = toString.head.toLower
+
+object Segment:
+
+  val fromChar = values.map(s => s.char -> s).toMap
```
julienrf, line 105: Here too, the type of `fromChar` would be welcome.

src/day8.scala
```diff
@@ -0,0 +1,130 @@
...
+
+  def parseSeq(s: String): Segments =
+    s.map(fromChar)
+
```
julienrf, line 109:
```suggestion
end Segment
```

src/day8.scala
```diff
@@ -0,0 +1,130 @@
...
+      display <- input.linesIterator.map(getDisplay)
+      segments <- display.split(" ")
+      uniqueDigit <- parseUniqueDigit(segments)
+    yield
+      uniqueDigit
+
+  uniqueDigits.length
+end part1
+
+def part2(input: String) =
```
julienrf, line 37: I would prefer to see the implementation details of the part 1 (the method `parseUniqueDigit`) rather than having to scroll over the `part2`.

julienrf: Also, the return type would be welcome

src/day8.scala
```diff
@@ -0,0 +1,130 @@
...
+def readInput(): String =
+  Using.resource(Source.fromFile("input/day8"))(_.mkString)
+
+type Segments = Seq[Segment]
+
+def part1(input: String) =
+
+  def getDisplay(line: String) = line.split("[|]")(1).trim
+
+  val uniqueDigits =
```
julienrf, line 26: Type would be welcome too

src/day8.scala
```diff
@@ -0,0 +1,130 @@
...
+
+  val solutions = problems.map((cypher, plaintext) =>
+    plaintext.map(substitutions(cypher))
+  )
+
+  solutions.map(digitsToInt).sum
+
+end part2
+
+val parseUniqueDigit = Segment.parseSeq andThen Digit.lookupUnique
```
julienrf, line 59: Type would be welcome

src/day8.scala
```diff
@@ -0,0 +1,130 @@
...
+
+def substitutions(cypher: Seq[Segments]): Segments => Digit =
+
+  def lookup(section: Seq[Segments], withSegments: Set[Segment]): (Segments, Seq[Segments]) =
+    val found = section.find(s => (s.toSet & withSegments) == withSegments).get
```
julienrf, line 64: If we transform the segments to a set, do we want to model them as a Seq in the first place? (ie, is order important? uniqueness? I didn’t read the problem statement)

julienrf: Also, it is not clear why  `.get` is always expected to succeed, maybe there are preconditions for `withSegments` that would be worth indicating?

src/day8.scala
```diff
@@ -0,0 +1,130 @@
...
+  val one = uniques(One)
+  val four = uniques(Four)
+  val seven = uniques(Seven)
+  val eight = uniques(Eight)
+  val (three, remainingFives) = lookup(ofSizeFive, withSegments = one.toSet)
+  val (nine, remainingSixes) = lookup(ofSizeSix, withSegments = three.toSet)
+  val (zero, Seq(six)) = lookup(remainingSixes, withSegments = seven.toSet)
+  val (five, Seq(two)) = lookup(remainingFives, withSegments = six.toSet &~ (eight.toSet &~ nine.toSet))
+
+  val index = Seq(zero, one, two, three, four, five, six, seven, eight, nine)
```
julienrf, line 87: Type would be welcome

src/day8.scala
```diff
@@ -0,0 +1,130 @@
...
+val parseUniqueDigit = Segment.parseSeq andThen Digit.lookupUnique
+
+def substitutions(cypher: Seq[Segments]): Segments => Digit =
+
+  def lookup(section: Seq[Segments], withSegments: Set[Segment]): (Segments, Seq[Segments]) =
+    val found = section.find(s => (s.toSet & withSegments) == withSegments).get
+    val remaining = section.filterNot(_ == found)
+    (found, remaining)
+
+  val uniques = Map.from(
```
julienrf, line 68: Type would be welcome

julienrf (commented):
Good job Jamie, I have left a few minor comments regarding the style.

src/day8.scala
```diff
@@ -0,0 +1,130 @@
...
+
+  val uniqueDigits =
+    for
+      display <- input.linesIterator.map(getDisplay)
+      segments <- display.split(" ")
+      uniqueDigit <- parseUniqueDigit(segments)
+    yield
+      uniqueDigit
+
+  uniqueDigits.length
```
sjrd, line 73: `length` is usually reserved for `Array`s and `String`s, for which it is the native method name. Consider using `size` instead on general Scala collections.

src/day8.scala
bishabosha: I moved `parseUniqueDigit` into `part1`

src/day8.scala
```diff
@@ -0,0 +1,136 @@
...
+    values.groupBy(_.segments.size).collect { case k -> Array(d) => k -> d }
+
+  def lookupUnique(segments: Segments): Option[Digit] =
+    uniqueLookup.get(segments.size)
+
+end Digit
+
+def part1(input: String): Int =
+
+  def getDisplay(line: String): String = line.split("\\|")(1).trim
```
sjrd, line 60: This is a bit obscure because of the escaping of `|` in the regex. Consider
```suggestion
  def getDisplay(line: String): String = line.split('|')(1).trim
```
instead. With a `Char` argument, it is always interpreted literally, rather than as a regex.

src/day8.scala
```diff
@@ -0,0 +1,136 @@
...
+  uniqueDigits.length
+end part1
+
+def part2(input: String): Int =
+
+  def parseSegmentsSeq(segments: String): Seq[Segments] =
+    segments.trim.split(" ").toSeq.map(Segment.parseSeq)
+
+  def splitParts(line: String): (Seq[Segments], Seq[Segments]) =
+    val Array(cypher, plaintext) = line.split("\\|").map(parseSegmentsSeq)
```
sjrd, line 82: Same here:
```suggestion
    val Array(cypher, plaintext) = line.split('|').map(parseSegmentsSeq)
```

src/day8.scala
```diff
@@ -0,0 +1,136 @@
...
+
+  val char = toString.head.toLower
+
+object Segment:
+  type Segments = Seq[Segment]
+
+  val fromChar: Map[Char, Segment] = values.map(s => s.char -> s).toMap
+
+  def parseSeq(s: String): Segments =
+    s.map(fromChar)
```
sjrd, line 32: Why is this called parse*Seq*? It doesn't seem to be related to `Seq` at all. Consider using a name that reflects the semantics more than the types involved. (The types are there to talk about the types.)

src/day8.scala
```diff
@@ -0,0 +1,136 @@
...
+    uniqueLookup.get(segments.size)
+
+end Digit
+
+def part1(input: String): Int =
+
+  def getDisplay(line: String): String = line.split("\\|")(1).trim
+
+  val parseUniqueDigit: String => Option[Digit] =
+    Segment.parseSeq andThen Digit.lookupUnique
```
sjrd, line 63: It may be just me, but I find this usage of functions-as-methods and conversely quite disorientating. Have you considered instead:
```suggestion
  def parseUniqueDigit(s: String): Option[Digit] =
    Digit.lookupUnique(Segment.parseSeq(s))
```
?

src/day8.scala
```diff
@@ -0,0 +1,136 @@
...
+  )
+
+  solutions.map(digitsToInt).sum
+
+end part2
+
+def substitutions(cypher: Seq[Segments]): Segments => Digit =
+
+  def lookup(section: Seq[Segments], withSegments: Set[Segment]): (Segments, Seq[Segments]) =
+    val found = section.find(s => (s.toSet & withSegments) == withSegments).get
```
sjrd, line 101: Consider using `subsetOf`:
```suggestion
    val found = section.find(s => withSegments.subsetOf(s)).get
```

src/day8.scala
```diff
@@ -0,0 +1,136 @@
...
+    val remaining = section.filterNot(_ == found)
```
sjrd, line 102: Consider using `partition` instead:
```suggestion
    val (Seq(uniqueMatch), remaining) = section.partition(s => withSegments.subsetOf(s))
    (uniqueMatch, remaining)
```
This also has the advantage of failing early if there are two or more inputs matching the given segments.

src/day8.scala
sjrd: > If we transform the segments to a set, do we really want to model them as a Seq in the first place? (ie, is order important? uniqueness? I didn’t read the problem statement)

I agree that it would be nicer to model them as a `Set` from the start.

bishabosha: I had some trouble when I was implementing when I tried to use set directly, but now I was able to refactor to only use set, will push soon

src/day8.scala
bishabosha: thanks, this was very useful to know

src/day8.scala
bishabosha: changing to `parseSegments`

src/day8.scala
bishabosha: that is fair, I will use the suggestion

src/day8.scala
bishabosha: nice, thank you

src/day8.scala
bishabosha: I like this!

src/day8.scala
bishabosha: `lookup` expects that `withSegments` should be a unique subset of one of the elements of `section`, this works with the logic for this specific problem

sjrd (approved):

PR #48, opened 2021-12-09
Add solution for day 9

vincenzobaz:


src/day9.scala
```diff
@@ -0,0 +1,112 @@
...
+end Heightmap
+
+
+def drawGrid(height: Int, width: Int, data: (Position, Height)*) =
+  val provided: Map[Position, Height] = data.toMap
+  var x = 0
+  var y = 0
+  val result = StringBuilder()
+  while (y < height) do
+    while (x < width) do
```
sjrd, line 68: Hum ... Why not
```suggestion
  for y <- 0 until height do
    for x <- 0 until width do
```
and get rid of the increments and assignments?

src/day9.scala
```diff
@@ -0,0 +1,112 @@
...
+@main def part2(): Unit =
+  println(s"The solution is ${part2(readInput())}")
+
+def readInput(): String =
+  Using.resource(Source.fromFile("input/day9"))(_.mkString)
+
+type Height = Int
+case class Position(x: Int, y: Int)
+
+class Heightmap(val width: Int, val height: Int, private val data: Array[Height]):
```
sjrd, line 21: Consider using an `IArray`, if possible.

Also, what's the rationale for using a flat array, as opposed to an `Array[Array[Height]]`?

src/day9.scala
```diff
@@ -0,0 +1,112 @@
...
+
+  def neighborsOf(coord: Position): Iterator[(Position, Height)] =
+    val Position(x, y) = coord
+    val idx = coord.y * width + coord.x
+    Iterator(
+      Option.when(x > 0)(Position(x - 1, y) -> data(idx - 1)),
+      Option.when(x < width - 1)(Position(x + 1, y) -> data(idx + 1)),
+      Option.when(y > 0)(Position(x, y - 1) -> data(idx - width)),
+      Option.when(y < height - 1)(Position(x, y + 1) -> data(idx + width))
+    ).flatMap(Iterator.from)
```
sjrd, line 33: Here you're directly accessing `data` with flat indices, even though you defined an abstraction with `Position` as `apply`. You may want to reuse it. Perhaps like this:
```suggestion
    Iterator(
      Option.when(x > 0)(Position(x - 1, y)),
      Option.when(x < width - 1)(Position(x + 1, y)),
      Option.when(y > 0)(Position(x, y - 1)),
      Option.when(y < height - 1)(Position(x, y + 1))
    ).flatMap(Iterator.from).map(coord => coord -> apply(coord))
```

src/day9.scala
```diff
@@ -0,0 +1,112 @@
...
+  result.toString
+
+def part1(input: String): Int =
+  val heightMap = Heightmap.fromString(input)
+
+  heightMap.lowPointsPositions.map(heightMap(_) + 1).sum
+end part1
+
+
+def bassin(lowPoint: Position, heightMap: Heightmap): Set[Position] =
```
sjrd, line 84: Typo ;)
```suggestion
def basin(lowPoint: Position, heightMap: Heightmap): Set[Position] =
```

src/day9.scala
```diff
@@ -0,0 +1,112 @@
...
+  iter(Set.empty, toVisit, Set(lowPoint))
+
+
+def part2(input: String): Int =
+  val heightMap = Heightmap.fromString(input)
+  val lowPoints = heightMap.lowPointsPositions
+  val bassins = lowPoints.map(bassin(_, heightMap))
+
+  bassins
+    .map(_.size)
```
sjrd, line 106: Uh oh. `basins` is a `Set[Position]`. So the result of calling `.map` will be a `Set[Int]`. That means that basins of the same size will *collapse* as if there was only one. This can cause an issue if some of the 3 largest basins have the same size. You should turn `basins` into an ordered sequence before calling `.map`.

src/day9.scala
```diff
@@ -0,0 +1,112 @@
...
+  heightMap.lowPointsPositions.map(heightMap(_) + 1).sum
+end part1
+
+
+def bassin(lowPoint: Position, heightMap: Heightmap): Set[Position] =
+  @scala.annotation.tailrec
+  def iter(visited: Set[Position], toVisit: Queue[(Position, Height)], bassinAcc: Set[Position]): Set[Position] =
+    if toVisit.isEmpty then bassinAcc
+    else
+      val ((currentPos, currentValue), others) = toVisit.dequeue
```
sjrd, line 89:
```suggestion
      val ((currentPos, currentHeight), others) = toVisit.dequeue
```

src/day9.scala
```diff
@@ -0,0 +1,112 @@
...
+      val newNodes = heightMap.neighborsOf(currentPos).toList.filter { (pos, height) =>
+        !visited(currentPos) && height != 9 && currentValue < height
```
sjrd, line 91: Given that basins are always separated by `9`'s (which seems weird to me, but the statement is quite clear about that), I don't think `currentValue < height` matters here. I believe it can be removed without losing correctness.

That means that it is not necessary to store `currentValue` at all, and that therefore a `Queue[Position]` would be enough.

src/day9.scala
julienrf, line 33:
```suggestion
    ).flatten
```
?

sjrd (commented):
The algorithm definitely looks good. I have a few comments.

julienrf (commented):
I am surprised by the use of `Array` and `Iterator`. These things are mutable and require a lot more vigilance than immutable data structures. Is there a reason not to use immutable data structures?

src/day9.scala
vincenzobaz: I can do that, I do not have strong opinions on this. The function was used only to debug. Maybe I could even remove it

src/day9.scala
vincenzobaz: Absolutely, thanks! I always forget `IArray`.
Concerning the nested or flat data, maybe it is more understandable for readers to use the `Array[Array[_]]`.
Do you think it would be better?

sjrd: It's not necessarily better, as long as you're consistent. Currently your various methods sometimes address things by coordinates, sometimes by indices, so it's a bit hard to follow. If you always work with indices in a flat array, it's fine. If you always go through your `apply` method, which abstracts the flat array away, it's good too. If you have an `Array[Array[_]]`, it's good too. You just need to pick one and stick to it. ;)

src/day9.scala
vincenzobaz: I am not a big fan of `flatten` with different types.

vincenzobaz:
@julienrf do you think it would be better to use `Vector` instead of `Array`?

I use `Iterator` only to loop over data structures without the need to allocate intermediate lists or arrays. Should I replace with `List`s?

src/day9.scala
vincenzobaz: I tested your suggestion and you seem to be correct, although this is not intuitive for me.

You are right that `9` should not be included, however in my reasoning, I needed to include only elements that were bigger when *growing* the basin.
In other words `1 2 3 2`  I should stop at not include the `2` which led me to include this check.

I am a bit confused: do we extend until we reach a 9 no matter what value is encountered?

julienrf:
My preference is that unless it makes a significant difference, it is fine to use the “non-lazy” collection types. That being said, if you don’t want to create collections for your intermediate results, you can use `LazyList`, which does not suffer from the issues of `Iterator` (although it has its own issues as well, such as the possibility of creating memory leaks if you are not vigilant…). I would also just use `IndexedSeq` instead of `Array`, unless you need mutability.

src/day9.scala
sjrd: The thing is that the statement rules out such situations, because it says:

> Locations of height 9 do not count as being in any basin, and all other locations will always be part of exactly one basin.

So the `3` would either have to part of the basin on the left, or on the right. But neither both or none. That means you could never know on which side it is supposed to go, and that therefore such situations are illegal.

vincenzobaz: So when we consider all basins, all values will covered but 9.
It works looking at the example.
I am still puzzled but I understand better

sjrd (approved):

PR #53, opened 2021-12-10
Day 10 solution

vincenzobaz:


src/day10.scala
```diff
@@ -0,0 +1,118 @@
...
+  def isOpen: Boolean = state == SymbolState.Open
+
+  def opens(that: Symbol): Boolean =
+    import SymbolState.*
+    (this, that) match
+      case (Parenthesis(Open), Parenthesis(Close)) => true
+      case (Bracket(Open), Bracket(Close)) => true
+      case (Brace(Open), Brace(Close)) => true
+      case (Diamond(Open), Diamond(Close)) => true
+      case _ => false
```
julienrf, line 49: I wonder if the kind of symbol (`Parenthesis`, `Bracket`, etc.) should be a data of the model instead of being a “constructor”. Then `opens` would just be like so:

```suggestion
  def opens(that: Symbol): Boolean = this.kind == that.kind
```

src/day10.scala
```diff
@@ -0,0 +1,118 @@
...
+  @scala.annotation.tailrec
+  def iter(pending: List[Symbol], input: LazyList[Symbol]): CheckResult =
+    if input.isEmpty then
+      if pending.isEmpty then CheckResult.Ok
+      else CheckResult.Incomplete(pending)
+    else
+      val (nextChar #:: remainingChars) = input 
+      if nextChar.isOpen then iter(nextChar :: pending, remainingChars)
+      else pending match
+        case Nil => CheckResult.IllegalClosing(None, nextChar)
```
vincenzobaz, line 61: This case never occurs with the provided input. I am not sure whether the handout says it cannot happen, but I prefer to have an exhaustive check

src/day10.scala
julienrf: Basically, you would model an opening/closing symbol like so:

~~~ scala
enum Direction:
  case Open, Close
enum Kind:
  case Parenthesis, Bracket, Brace, Diamond
case class Symbol(kind: Kind, direction: Direction)
~~~

vincenzobaz, line 49: This encoding might not be optimal

src/day10.scala
```diff
@@ -0,0 +1,118 @@
...
+      case Diamond(_) => 25137
+
+enum SymbolState:
+  case Open, Close
+
+enum Symbol(val state: SymbolState):
+  case Parenthesis(override val state: SymbolState) extends Symbol(state)
+  case Bracket(override val state: SymbolState) extends Symbol(state)
+  case Brace(override val state: SymbolState) extends Symbol(state)
+  case Diamond(override val state: SymbolState) extends Symbol(state)
```
julienrf, line 38: This syntax looks complicated to me.

I wonder if the following works:

```suggestion
enum Symbol:
  def state: SymbolState
  case Parenthesis(state: SymbolState)
  case Bracket(state: SymbolState)
  case Brace(state: SymbolState)
  case Diamond(state: SymbolState)
```

If it doesn’t, just fallback to a trait:

```suggestion
sealed trait Symbol:
  def state: SymbolState
case Parenthesis(state: SymbolState) extends Symbol
case Bracket(state: SymbolState) extends Symbol
case Brace(state: SymbolState) extends Symbol
case Diamond(state: SymbolState) extends Symbol
```

Both suggestion look lighter to my eyes.

src/day10.scala
```diff
@@ -0,0 +1,118 @@
...
+  def opens(that: Symbol): Boolean =
+    import SymbolState.*
+    (this, that) match
+      case (Parenthesis(Open), Parenthesis(Close)) => true
+      case (Bracket(Open), Bracket(Close)) => true
+      case (Brace(Open), Brace(Close)) => true
+      case (Diamond(Open), Diamond(Close)) => true
+      case _ => false
+
+def checkChunks(expression: LazyList[Symbol]): CheckResult =
```
julienrf, line 51: Not sure what motivates the use of a `LazyList` here as opposed to a basic `List`.

julienrf (approved):
Nice! I find the algorithm easy to read! I have left a few minor comments related to code style. Feel free to ignore them, they are just suggestions!

src/day10.scala
vincenzobaz: I like this encoding. Simpler is better

src/day10.scala
vincenzobaz: In theory we might reach an `IllegalClosing` before arriving at the end of the line. In this case there is no need to load the entirety of the line.
However I am not sure that this has any significative impact

julienrf: How large is the list? You already load the whole input in memory as a `String`, for the record.

src/day10.scala
```diff
@@ -0,0 +1,111 @@
...
+    if input.isEmpty then
+      if pending.isEmpty then CheckResult.Ok
+      else CheckResult.Incomplete(pending)
+    else
+      val (nextChar #:: remainingChars) = input 
+      if nextChar.isOpen then iter(nextChar :: pending, remainingChars)
+      else pending match
+        case Nil => CheckResult.IllegalClosing(None, nextChar)
+        case lastOpened :: previouslyOpened =>
+          if lastOpened.opens(nextChar) then iter(previouslyOpened, remainingChars)
```
julienrf, line 56: Here we already know that `lastOpened`’s direction is `Open`, and that `nextChar`’s direction is `Close`. So, we only need to check whether `lastOpened.kind == nextChar.kind` (instead of calling `lastOpened.opens(nextChar)`).

sjrd (approved):

vincenzobaz:
Thank you @sjrd and @julienrf  for your attentive reviews!

PR #38, opened 2021-12-07
Add the blog post about the solution for day 7

tgodzik:

PR #57, opened 2021-12-10
Day 8 article

bishabosha:


julienrf:
Don’t we need to also do something like this? https://github.com/scalacenter/scala-advent-of-code/pull/38/files#diff-ac42b5536aba05e65aa12e65c4f44017645c48c18c9dd4893b93da1d2a1acf28

bishabosha:
> Don’t we need to also do something like this? https://github.com/scalacenter/scala-advent-of-code/pull/38/files#diff-ac42b5536aba05e65aa12e65c4f44017645c48c18c9dd4893b93da1d2a1acf28

All days have been added now so it does not need to adjust the homepage

docs/puzzles/day8.md
```diff
@@ -2,10 +2,616 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+## Solution of Part 1
+
+### Modelling the Domain
+
+First we will model our problem and parse the input into our model.
+
+#### `Segment`
+
+As there are a fixed range of possible display _segments_:
+`a`, `b`, `c`, `d`, `e`, `f`, `g`, we will model them with an enumeration.
```
julienrf, line 20: I suggest introducing an ASCIIART picture showing the 8 segments, like you did in part 2.

docs/puzzles/day8.md
```diff
@@ -6,6 +6,610 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+#### 5 and 6 Segment Sections
+
+We will create sections of encoded digits of size 5 and size 6:
+
+```scala
+val ofSizeFive = cipher.filter(encoded => encoded.sizeIs == 5)
+val ofSizeSix = cipher.filter(encoded => encoded.sizeIs == 6)
+```
+> `filter` is a method defined in the `Seq` class that takes a function argument (evaluated on every element).
+> It will create a new list from `cipher` where elements that do not match its argument are dropped.
```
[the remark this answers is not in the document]
bishabosha: good point :)

julienrf (approved):
This is an amazing write-up!

docs/puzzles/day8.md
bishabosha: I added a diagram at the start, what do you think?

julienrf: This is perfect!

sjrd (approved):

PR #63, opened 2021-12-12
small change to URL for day 1 solution

philipschwarz:
point to a dir rather than to a file, so it can change name without breaking the URL


tgodzik (approved):
LGTM!

PR #64, opened 2021-12-12
Update day1.md

justinhj:
Added my solution to day 1


tgodzik (approved):
Thanks! This is awesome to see so many different solutions here!

PR #66, opened 2021-12-13
Add article for day 4 by @Sporarum

adpi2:


docs/puzzles/day4.md
```diff
@@ -1,11 +1,232 @@
...
+2. for each `line` in `lines` put `parseLine(line)` in the list
+
+This gives us a `List[List[Int]]`, which we use to construct a `Board`.
+
+Since we have multiple `Board`s:
+```scala
+val originalBoards: List[Board] = inputSections.tail.map(Board.parse)
+```
+
+## Reasonning about the problem
```
bishabosha, line 119:
```suggestion
## Reasoning about the problem
```

docs/puzzles/day4.md
```diff
@@ -1,11 +1,232 @@
...
+It's kind of difficult to think about all those numbers being picked at random turn.
+We can simplify the problem by replacing each number by the "turn" at which it was drawn.
+
+`zipWithIndex` transforms our list of numbers into a list of number-index pair, where the index is just the turn at which the number is picked.
+We can then convert it to a `Map`, to be able to use it like a function.
+To be able also to go back, we invert our `Map` by swapping its keys and values.
+
+```scala
+val numberToTurn = numbers.zipWithIndex.toMap
+val turnToNumber = numberToTurn.map((number -> turn) => (turn -> number))
```
bishabosha, line 130: this isn't valid syntax
```suggestion
val turnToNumber = numberToTurn.map { case (number -> turn) => (turn -> number) }
```
```suggestion
val turnToNumber = numberToTurn.map((number, turn) => turn -> number)
```

docs/puzzles/day4.md
```diff
@@ -1,11 +1,232 @@
...
+### Parsing Numbers
+
+For the numbers, since they are separated by `,` and nothing else, we can parse them with:
+```scala
+val numbers: List[Int] = inputSections.head.split(',').map(_.toInt)
+```
+
+### Parsing Boards
+
+A board is just a table of integers, and a table is just a list of lines, where each line is also list.
```
sjrd, line 67:
```suggestion
A board is a table of integers, and a table is a list of lines, where each line is also list.
```

docs/puzzles/day4.md
```diff
@@ -1,11 +1,232 @@
...
+And so we have our type for the boards: `List[List[Int]]`!
+
+But not so fast, we would like to add some extra operations on boards, so we wrap it in a [case classe](https://docs.scala-lang.org/tour/case-classes.html):
```
sjrd, line 70:
```suggestion
But not so fast, we would like to add some extra operations on boards, so we wrap it in a [case class](https://docs.scala-lang.org/tour/case-classes.html):
```

docs/puzzles/day4.md
```diff
@@ -1,11 +1,232 @@
...
+```
+
+Now that we have our representation, we still have to actually parse the string that represents a board, for that we'll create a [companion object](https://docs.scala-lang.org/scala3/book/taste-objects.html) with a method `parse` that takes a `String` as input, and returns a `Board`:
+
+```scala
+object Board:
+  def parse(inputBoard: String): Board = ???
+```
+
+Let's start easy, assuming we have a line, how do we find all the numbers?
```
sjrd, line 85:
```suggestion
Let's start from the ground up. Assuming we have a line, how do we find all the numbers?
```

docs/puzzles/day4.md
```diff
@@ -1,11 +1,232 @@
...
+We can use a [regular expressions](https://en.wikipedia.org/wiki/Regular_expression)(Regexes):
```
sjrd, line 86:
```suggestion
We can use a [regular expression](https://en.wikipedia.org/wiki/Regular_expression)(Regexes): 
```

docs/puzzles/day4.md
```diff
@@ -1,11 +1,232 @@
...
+```
+`raw` allows us to write things like `\d` without it being translated as a line return.
+`.r` converts our `String` to a `Regex`.
+
+For each line, `numberParser` finds every number in it, and we parse them to `Int`:
+```scala
+def parseLine(inputLine: String): List[Int] =
+  numberParser.findAllIn(inputLine).toList.map(_.toInt)
+```
+And the lines are just separated by newlines:
```
sjrd, line 102:
```suggestion
And the lines are separated by newlines:
```

docs/puzzles/day4.md
```diff
@@ -1,11 +1,232 @@
...
+```scala
+val originalBoards: List[Board] = inputSections.tail.map(Board.parse)
+```
+
+## Reasonning about the problem
+
+It's kind of difficult to think about all those numbers being picked at random turn.
+We can simplify the problem by replacing each number by the "turn" at which it was drawn.
+
+`zipWithIndex` transforms our list of numbers into a list of number-index pair, where the index is just the turn at which the number is picked.
```
sjrd, line 124:
```suggestion
`zipWithIndex` transforms our list of numbers into a list of number-index pair, where the index, in this case, is the turn at which the number is picked.
```

docs/puzzles/day4.md
```diff
@@ -1,11 +1,232 @@
...
+To be able also to go back, we invert our `Map` by swapping its keys and values.
+
+```scala
+val numberToTurn = numbers.zipWithIndex.toMap
+val turnToNumber = numberToTurn.map((number -> turn) => (turn -> number))
+```
+
+Our simplified boards are therefore:
+```scala
+val boards = originalBoards.map(board => board.mapNumbers(numberToTurn))
```
sjrd, line 135: You said earlier that you would explain what `mapNumbers` is, but you haven't done so.

docs/puzzles/day4.md
```diff
@@ -1,11 +1,232 @@
...
+```
+
+We still need to do one more thing before we can solve the problem: computing the score of a board.
+The score is the sum of all numbers that have not been drawn yet, times the turn at which the board wins.
+
+```scala
+def score(board: Board, turn: Int) = ???
+```
+
+For each line, the numbers that have not been drawn are simply the ones bigger than the winning turn of that board.
```
sjrd, line 172:
```suggestion
For each line, the numbers that have not been drawn are the ones bigger than the winning turn of that board.
```

docs/puzzles/day4.md
```diff
@@ -1,11 +1,232 @@
...
+We filter them with `lines.filter(_ > turn)`.
+
+However, simply taking the sum would be wrong, as we are using the turns, and not the original numbers!
```
sjrd, line 175:
```suggestion
However, only taking the sum would be wrong, as we are using the turns, and not the original numbers!
```

docs/puzzles/day4.md
```diff
@@ -1,11 +1,232 @@
...
+We thus need to map them to their original values:
+```scala
+val sumNumsNotDrawn = board.lines.map{ line => 
+  line.filter(_ > turn).map(turnToNumber(_)).sum
+}.sum
+```
+
+The score is then simply:
```
sjrd, line 183:
```suggestion
The score is then:
```

sjrd (commented):
There are a bunch of "easy words" to remove.

PR #67, opened 2021-12-13
Add solution of day13

adpi2:
https://adventofcode.com/2021/day/13


src/day13.scala
```diff
@@ -0,0 +1,66 @@
...
+def readInput(): String =
+  Using.resource(Source.fromFile("input/day13"))(_.mkString)
+
+def part1(input: String): Int =
+  val (dots, folds) = parseInstructions(input)
+  dots.map(folds(0).apply).size
+
+def part2(input: String): String =
+  val (dots, folds) = parseInstructions(input)
+  val foldedDots = folds.foldLeft(dots)((dots, fold) => dots.map(fold.apply))
```
vincenzobaz, line 25: `dots` is shadowed in the `foldLeft`. You might consider another variable name

src/day13.scala
```diff
@@ -0,0 +1,66 @@
...
+  val (length, height) = (foldedDots.map(_.x).max + 1, foldedDots.map(_.y).max + 1)
+  val paper = Array.fill(height, length)('.')
+  for dot <- foldedDots do paper(dot.y)(dot.x) = '#'
+  
+  paper.map(_.mkString).mkString("\n")
+
+def parseInstructions(input: String): (Set[Dot], List[Fold]) =
+  val sections = input.split("\n\n")
+  val dots = sections(0).split('\n').toSet.map(Dot.parse)
+  val folds = sections(1).split('\n').toList.map(Fold.parse)
```
vincenzobaz, line 36: `linesIterator.map(...).toSet` might be more efficient

src/day13.scala
```diff
@@ -0,0 +1,66 @@
...
+  (dots, folds)
+
+def fold(along: Int)(point: Int): Int =
```
vincenzobaz, line 39: `fold` could be `private` to `enum Fold`.

src/day13.scala
```diff
@@ -0,0 +1,66 @@
...
+@main def part2(): Unit =
+  val answer = part2(readInput())
+  println(s"The answer is:\n$answer")
+
+def readInput(): String =
+  Using.resource(Source.fromFile("input/day13"))(_.mkString)
+
+def part1(input: String): Int =
+  val (dots, folds) = parseInstructions(input)
+  dots.map(folds(0).apply).size
```
vincenzobaz, line 21: Maybe `head` is more idiomatic on lists than `(0)`

vincenzobaz (approved):
It looks great! Just a few remarks

src/day13.scala
adpi2: :+1:

src/day13.scala
```diff
@@ -0,0 +1,66 @@
...
+
+def parseInstructions(input: String): (Set[Dot], List[Fold]) =
+  val sections = input.split("\n\n")
+  val dots = sections(0).split('\n').toSet.map(Dot.parse)
+  val folds = sections(1).split('\n').toList.map(Fold.parse)
+  (dots, folds)
+
+def fold(along: Int)(point: Int): Int =
+  if point < along then point
+  else 2 * along - point
```
sjrd, line 41: My intuition here would be to write `along - (point - along)`. Of course these are equivalent, but I can only convince myself that `2 * along - point` is correct by verifying that it is equivalent to `along - (point - along)`. Perhaps add a comment or something?

src/day13.scala
adpi2: Unfortunately I cannot find a better name.

julienrf: `foldedDotsAcc`? Not sure it is better :laughing:

src/day13.scala
```diff
@@ -0,0 +1,67 @@
...
+
+def part2(input: String): String =
+  val (dots, folds) = parseInstructions(input)
+  val foldedDots = folds.foldLeft(dots)((dots, fold) => dots.map(fold.apply))
+  
+  val (width, height) = (foldedDots.map(_.x).max + 1, foldedDots.map(_.y).max + 1)
+  val paper = Array.fill(height, width)('.')
+  for dot <- foldedDots do paper(dot.y)(dot.x) = '#'
+  
+  paper.map(_.mkString).mkString("\n")
```
julienrf, line 31: Seems beautiful, I want to see the result!

src/day13.scala
```diff
@@ -0,0 +1,67 @@
...
+  val dots = sections(0).linesIterator.map(Dot.parse).toSet
+  val folds = sections(1).linesIterator.map(Fold.parse).toList
+  (dots, folds)
+
+case class Dot(x: Int, y: Int)
+
+object Dot:
+  def parse(line: String): Dot =
+    line match
+      case s"$x,$y" => Dot(x.toInt, y.toInt)
```
julienrf, line 44: :star_struck:

PR #68, opened 2021-12-13
Add article for day 13

adpi2:


docs/puzzles/day13.md
```diff
@@ -1,11 +1,162 @@
...
+```scala
+case class Dot(x: Int, y: Int)
+
+enum Fold:
+  case Vertical(x: Int)
+  case Horizontal(y: Int)
+```
+
+A `Dot` is made of two integers: `x` and `y`.
+A `Fold` is either `Vertical` if it has an `x` coordinate or `Horizontal` if it has an `y` coordinate.
```
sjrd, line 24:
```suggestion
A `Fold` is either `Vertical` if it has an `x` coordinate or `Horizontal` if it has a `y` coordinate. 
```
**an** ex (x) but **a** why (y)

docs/puzzles/day13.md
```diff
@@ -1,11 +1,162 @@
...
+
+```scala
+def fold(along: Int)(value: Int): Int =
+  if value < along then value
+  else along - (value - along)
+```
+
+We can check this formula with some examples:
+ - Folding a dot at 2 along 7 does not move the dot because `2 < 7`.
+ - Folding a dot at 9 along 7 (`......|.#`) move the dot to `7 - 2 = 5` (`....#.|..`) because `9 = 7 + 2`.
```
sjrd, line 97: Perhaps with less words:
```suggestion
 - Folding a dot at 9 along 7 (`......|.#`) moves the dot to `7 - (9 - 7) = 7 - 2 = 5` (`....#.|..`).
```
?

sjrd (approved):

mlachkar (approved):

PR #71, opened 2021-12-14
Solution for day 14

sjrd:


src/day14.scala
```diff
@@ -0,0 +1,112 @@
...
+  val (initialPolymer, insertionRules) = parseInput(input)
+
+  /* S(ab...f, n) :=
+   *   frequencies of all letters *except the first one* in the expansion of "ab...f" after n iterations
+   *
+   * Frequencies are multisets. We use + and ∑ to denote their sum.
+   *
+   * For any string longer than 2 chars, we have
+   *   S(x₁x₂x₃...xₚ, n) = ∑(S(xᵢxⱼ, n)) for i = 0 to n-1 and j = i+1
+   * because each initial pair expands independently of the others. Each initla
```
julienrf, line 61:
```suggestion
   * because each initial pair expands independently of the others. Each initial
```

src/day14.scala
```diff
@@ -0,0 +1,112 @@
...
+   * And that means we can iteratively construct S(xy, n) for all pairs 'xy'
+   * and for n ranging from 0 to 40.
+   */
+
+  // S : (charPair, n) -> frequencies of everything but the first char after n iterations from charPair
+  val S = mutable.Map.empty[(CharPair, Int), Frequencies]
+
+  // Base case: S(xy, 0) = {y -> 1} for all x, y
+  for (pair, insert) <- insertionRules do
+    S((pair, 0)) = Map(pair._2 -> 1L)
```
julienrf, line 88:
```suggestion
    S((pair, 0)) = Map(pair(1) -> 1L)
```

(I like the homogeneity with collections)

julienrf (approved):
Brilliant.

I left only minor suggestions or typos.

src/day14.scala
```diff
@@ -0,0 +1,112 @@
...
+type Polymer = List[Char]
+type CharPair = (Char, Char)
+type InsertionRules = Map[(Char, Char), Char]
+type Frequencies = Map[Char, Long]
+
+def part1(input: String): Long =
+  val (initialPolymer, insertionRules) = parseInput(input)
+  val finalPolymer = (0 until 10).foldLeft(initialPolymer)((polymer, _) => applyRule(polymer, insertionRules))
+  val frequencies: Frequencies = finalPolymer.groupMapReduce(identity)(_ => 1L)(_ + _)
+  val max = frequencies.maxBy((char, count) => count)._2
```
julienrf, line 29: Here too, you could use the “index-like” access to the pair:

```suggestion
  val max = frequencies.maxBy((char, count) => count)(1)
```

There are other places like that below.

src/day14.scala
```diff
@@ -0,0 +1,112 @@
...
+def parseRule(line: String): (CharPair, Char) =
+  line match
+    case s"$pairStr -> $inserStr" => (pairStr(0), pairStr(1)) -> inserStr(0)
+    case _                        => throw new Exception(s"Cannot parse '$line' as an insertion rule")
+
+def applyRule(polymer: Polymer, rules: InsertionRules): Polymer =
+  val pairs = polymer.zip(polymer.tail)
+  val insertionsAndSeconds: List[List[Char]] =
+    for pair <- pairs
+      yield rules(pair) :: pair._2 :: Nil
```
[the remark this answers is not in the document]
sjrd: Is that really the recommended default that we should use all the time? I don't use Scala 3 much, but I still feel that the field access has a stronger feel.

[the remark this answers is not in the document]
julienrf: I am not sure, I just personally like the consistency with the collections API. After all, a tuple is a heterogeneous collection with a fixed size.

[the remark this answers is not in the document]
sjrd: Meh... I think I'll keep using `_2` unless more people think we should use `(1)` here.

src/day14.scala
```diff
@@ -0,0 +1,113 @@
...
+  // S(polymer, 40) = ∑(S(pair, 40))
+  val pairsInPolymer = initialPolymer.zip(initialPolymer.tail)
+  val polymerS = (for pair <- pairsInPolymer yield S(pair, 40)).reduce(addFrequencies)
+
+  // We have to add the very first char to get all the frequencies
+  val frequencies = addFrequencies(polymerS, Map(initialPolymer.head -> 1L))
+
+  // Finally, we can finish the computation as in part 1
+  val max = frequencies.maxBy((char, count) => count)._2
+  val min = frequencies.minBy((char, count) => count)._2
```
tgodzik, line 107: This gets rid of `_.2` altogether
```suggestion
  val max = frequencies.map((char, count) => count).max
  val min = frequencies.map((char, count) => count).min
```

src/day14.scala
```diff
@@ -0,0 +1,113 @@
...
+   */
+
+  // S : (charPair, n) -> frequencies of everything but the first char after n iterations from charPair
+  val S = mutable.Map.empty[(CharPair, Int), Frequencies]
+
+  // Base case: S(xy, 0) = {y -> 1} for all x, y
+  for (pair, insert) <- insertionRules do
+    S((pair, 0)) = Map(pair._2 -> 1L)
+
+  // Recursive case S(xy, n) = S(xz, n - 1) + S(zy, n - 1) with y = insertionRules(xy)
```
tgodzik, line 91:
```suggestion
  // Recursive case S(xy, n) = S(xz, n - 1) + S(zy, n - 1) with z = insertionRules(xy)
```

src/day14.scala
sjrd: Indeed. I even simplified to `frequencies.values.max`.

src/day14.scala
```diff
@@ -0,0 +1,113 @@
...
+def parseRule(line: String): (CharPair, Char) =
+  line match
+    case s"$pairStr -> $inserStr" => (pairStr(0), pairStr(1)) -> inserStr(0)
+    case _                        => throw new Exception(s"Cannot parse '$line' as an insertion rule")
+
+def applyRules(polymer: Polymer, rules: InsertionRules): Polymer =
+  val pairs = polymer.zip(polymer.tail)
+  val insertionsAndSeconds: List[List[Char]] =
+    for pair <- pairs
+      yield rules(pair) :: pair._2 :: Nil
```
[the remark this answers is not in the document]
sjrd: Good point. I also used this suggestion, and others getting rid of the `_2`.

src/day14.scala
```diff
@@ -0,0 +1,113 @@
...
+   *   S(xy, n)
+   *     = S(xzy, n-1) where z is the insertion char for the pair 'xy' (by definition)
+   *     = S(xz, n-1) + S(zy, n-1) -- the particular case of 3 chars above
+   *
+   * And that means we can iteratively construct S(xy, n) for all pairs 'xy'
+   * and for n ranging from 0 to 40.
+   */
+
+  // S : (charPair, n) -> frequencies of everything but the first char after n iterations from charPair
+  val S = mutable.Map.empty[(CharPair, Int), Frequencies]
```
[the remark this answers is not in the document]
sjrd: The mutable map is not even necessarily more efficient. It's possible to do this purely functionally.
But this is the way I could express my problem in the more straightforward way. I "build up" the set of `(CharPair, Int)` that I have computed. I don't want to "dive down" into the ones I would need to compute new ones.

This is the Scala way for me: use immutable stuff when that expresses the ideas better, and notably in-between components. But also use mutable stuff when that expresses the ideas better.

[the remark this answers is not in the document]
tgodzik: Sure, makes sense!

tgodzik (approved):
:rocket:

src/day14.scala
```diff
@@ -0,0 +1,113 @@
...
+  // Base case: S(xy, 0) = {y -> 1} for all x, y
+  for (pair @ (first, second), insert) <- insertionRules do
+    S((pair, 0)) = Map(second -> 1L)
+
+  // Recursive case S(xy, n) = S(xz, n - 1) + S(zy, n - 1) with z = insertionRules(xy)
+  for n <- 1 to 40 do
+    for (pair, insert) <- insertionRules do
+      val (x, y) = pair
+      val z = insertionRules(pair)
+      S((pair, n)) = addFrequencies(S((x, z), n - 1), S((z, y), n - 1))
```
tgodzik, line 96: Btw. This seems counter intuitive, so it would probably be good to explain it really well in the blog.

I kind of expect something like:
```
S(((x, z), n)) = addFrequencies(S(((x, z), n)) , S(pair, n - 1))
S(((z, y), n)) = addFrequencies(S(((z, y), n)) , S(pair, n - 1))
```
Such as n  `xy` produces n `xz` and n `zy` 

But obviously this works, but having issues wrapping my head around it.

sjrd: The big comment above does not clear it up for you? Anyway, I am of course spending a bit of space on that in the article.

tgodzik: After reading it very carefully it is much clearer, it's just I think normally instead of:
- `S(xy, n)` meaning it's what happens with xy after n steps
people might understand it as
- `S(xy, n)` meaning number of xy pairs after n steps

tgodzik: But anyway, I think the solution is super cool and great as a learning tool!

PR #72, opened 2021-12-14
Track people with ad-blockers

julienrf:
It seems that some ad-blockers prevent the script `plausible.js` to be downloaded. If we use the name `script.js` it works.


adpi2:
Nice! :)

PR #76, opened 2021-12-16
Add solution for day 16

tgodzik:


vincenzobaz (approved):

src/day16.scala
```diff
@@ -0,0 +1,201 @@
...
+/*
+ * Parsing of packets
+ */
+
+type BinaryData = List[Char] 
+
+inline def toInt(chars: BinaryData): Int =
+  Integer.parseInt(chars.mkString, 2)
+
+inline def toLong(chars: BinaryData): Long =
```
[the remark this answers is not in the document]
tgodzik: It's not, maybe I should remove it in order not to confuse people.

src/day16.scala
```diff
@@ -0,0 +1,201 @@
...
+
+case class LesserThanPacket(version: Int, lhs: Packet, rhs: Packet)
+    extends BinaryOperatorPacket:
+  val typeId = 6
+  def value = if lhs.value < rhs.value then 1 else 0
+
+case class EqualsPacket(version: Int, lhs: Packet, rhs: Packet)
+    extends BinaryOperatorPacket:
+  val typeId = 7
+  def value = if lhs.value == rhs.value then 1 else 0
```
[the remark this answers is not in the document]
tgodzik: Maybe? I will take a look.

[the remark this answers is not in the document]
tgodzik: Actually, not sure how to model it, since each case class has different data, the only stable thing is typeId, which I only added for clarity, it's not actually needed.

What would be your idea about it?

src/day16.scala
```diff
@@ -0,0 +1,201 @@
...
+
+sealed trait Packet:
+  def version: Int
+  def typeId: Int
+  def value: Long
+
+sealed trait OperatorPacket extends Packet:
+  def version: Int
+  def typeId: Int
+  def value: Long
```
adpi2, line 49: no need to repeat it here

src/day16.scala
```diff
@@ -0,0 +1,201 @@
...
+  def exprs: List[Packet]
+
+sealed trait BinaryOperatorPacket extends OperatorPacket:
+  def version: Int
+  def typeId: Int
+  def value: Long
```
adpi2, line 55: same here

src/day16.scala
```diff
@@ -0,0 +1,201 @@
...
+  def rhs: Packet
+  def lhs: Packet
+  def exprs = List(lhs, rhs)
```
adpi2, line 58: would it be more clear with a `override`?

src/day16.scala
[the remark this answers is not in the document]
vincenzobaz: My first reaction was something similar to

```scala
enum Packet(version: Int, typedId: Int, value: Int):
  case UnaryOperator(version: Int, typedId: Int, value: Int, exprs: List[Packet]) extends Packet(version, typedId, value)
  case BinaryOperator(version: Int, typedId: Int, value: Int, rhs: Packet, lhs: Packet) extends UnaryOperator(version, typedId, value, List(rhs, lhs))
```
although I am not sure it would work for the entire hierarchy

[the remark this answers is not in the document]
adpi2: What about something like:

```
enum Packet(version: Int):
  case Sum(version: Int, packets: List[Packet]) extends Packet(version)
  case Product(version: Int, packets: List[Packet]) extends Packet(version)
  ...

  def value: Int =
    this match
      case Sum(version, packets) => packets.map(_.value).sum
      case Product(version, packets) => packets.reduce(_ * _)
      ... 
```

src/day16.scala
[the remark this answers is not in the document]
sjrd: Yes, in general, it's better to reserve `inline` for things that would not *compile* without `inline` (for example, metaprogramming), or when there is a compelling reason to use it. If there is no compelling reason, it's better not to use `inline`.

src/day16.scala
```diff
@@ -0,0 +1,201 @@
...
+type BinaryData = List[Char] 
+
+inline def toInt(chars: BinaryData): Int =
+  Integer.parseInt(chars.mkString, 2)
+
+inline def toLong(chars: BinaryData): Long =
+  java.lang.Long.parseLong(chars.mkString, 2)
+
+@tailrec
+def readLiteralBody(tail: BinaryData, numAcc: BinaryData): (Long, BinaryData) =
```
sjrd, line 111: Sometimes you use `tail`, sometimes `current`, sometimes `packet`. Consider always using the same name to represent the input, for example `input` ;).

src/day16.scala
```diff
@@ -0,0 +1,201 @@
...
+      acc: List[Packet]
+  ): (List[Packet], BinaryData) =
+    if remaining == 0 then (acc, current)
+    else
+      val (newExpr, rest) = decodePacket(current)
+      readMaxPackages(rest, remaining - 1, acc :+ newExpr)
+
+  lenId match
+    // read based on length
+    case List('0') =>
```
sjrd, line 146: Have you considered
```scala
if lenId(0) == '0' then
```
like you did in `readLiteralBody`?

sjrd (approved):

src/day16.scala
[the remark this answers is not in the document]
tgodzik: Removed!

src/day16.scala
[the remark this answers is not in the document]
tgodzik: Changed! Look pretty nice, though summing up versions is a bit more verbose now.

src/day16.scala
tgodzik: Renamed! I haven't noticed the discrepancy when beautifying the code.

src/day16.scala
tgodzik: Changed, good idea. Wasn't too happy with that code.

src/day16.scala
```diff
@@ -0,0 +1,163 @@
...
+  case GreaterThan(version: Int, lhs: Packet, rhs: Packet) extends Packet(version, 5)
+  case LesserThan(version: Int, lhs: Packet, rhs: Packet) extends Packet(version, 6)
+  case Equals(version: Int, lhs: Packet, rhs: Packet) extends Packet(version, 7)
+
+  def versionSum: Int =
+    this match
+      case Sum(version, exprs)            => version + exprs.map(_.versionSum).sum
+      case Product(version, exprs)        => version + exprs.map(_.versionSum).sum
+      case Minimum(version, exprs)        => version + exprs.map(_.versionSum).sum
+      case Maximum(version, exprs)        => version + exprs.map(_.versionSum).sum
```
vincenzobaz, line 56: I think you can use `|` in pattern matches:
```
case x: Sum | Product | Minimum | Maximum
```
and 
```
case y: GreaterThabn | LesserThan | Equals
```

tgodzik: That doesn't work unfortunately:
```
[error] ./src/day16.scala:53:12: Illegal variable x in pattern alternative
[error]       case x : Sum | Product | Minimum| Maximum => x.version + x.exprs.map(_.versionSum).sum
[error]            ^^^^^^^
```

PR #80, opened 2021-12-17
solve day 17 parts 1 and 2

bishabosha:


sjrd (approved):
Looks good.

tgodzik (approved):
Looks good! And seem perfectly understandable. We could probably optimize it by iterating on y only, but that is for sure not necessary.

src/day17.scala
```diff
@@ -0,0 +1,77 @@
...
+
+def collides(p: Probe, t: Target): Boolean =
+  val Probe(Position(px, py), _) = p
+  val Target(xs, ys) = t
+  xs.contains(px) && ys.contains(py)
+
+def beyond(p: Probe, target: Target): Boolean =
+  val Probe(Position(px, py), Velocity(vx, vy)) = p
+  val Target(xs, ys) = target
+  val beyondX = vx == 0 && px < xs.min || px > xs.max
```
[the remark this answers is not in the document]
bishabosha: done, thanks!

src/day17.scala
```diff
@@ -0,0 +1,77 @@
...
+  beyondX || beyondY
+
+def simulate(probe: Probe, target: Target): Option[Int] =
+  LazyList
+    .iterate((probe, 0))((p, m) => (step(p), m `max` p.position.y))
+    .dropWhile((p, _) => !collides(p, target) && !beyond(p, target))
+    .headOption
+    .collect { case (p, m) if collides(p, target) => m }
+
+def run(target: Target)(positiveOnly: Boolean) =
```
julienrf, line 51:
```suggestion
def run(target: Target)(positiveOnly: Boolean): Seq[Int] =
```

Also, maybe a more self-explanatory name would be welcome, like `computeAllSolutions`?

julienrf (approved):

src/day17.scala
bishabosha: Thank you for the suggestion! there were a few inconsistent names so I updated them, I also replaced `dvx` in `step` by subtracting the sign of `vx`

PR #84, opened 2021-12-20
Add code for day 20

julienrf:


src/day20.scala
```diff
@@ -0,0 +1,114 @@
...
+      if image.pixel(xx, yy) == Pixel.Lit then
+        result = result | bitMask
+      end if
+      bitMask = bitMask << 1
+    result
+  end pixelValue
+
+end Enhancer
+
+class Image(pixels: IndexedSeq[IndexedSeq[Pixel]], val outboundsPixel: Pixel):
```
sjrd, line 81:
```suggestion
class Image(pixels: IndexedSeq[IndexedSeq[Pixel]], val outOfBoundsPixel: Pixel):
```

src/day20.scala
```diff
@@ -0,0 +1,114 @@
...
+    for
+      yy <- (y + 1) to (y - 1) by -1
+      xx <- (x + 1) to (x - 1) by -1
+    do
+      if image.pixel(xx, yy) == Pixel.Lit then
+        result = result | bitMask
+      end if
+      bitMask = bitMask << 1
+    result
+  end pixelValue
```
sjrd, line 77: It took me a moment to realize why you were going `by -1`. It's because you're shifting the bit mask to the left on every iteration.

Consider instead shifting the *result* to the left on each iteration, as follows:
```suggestion
  private def pixelValue(image: Image, x: Int, y: Int): Int =
    var result = 0
    for
      yy <- (y + 1) to (y - 1)
      xx <- (x + 1) to (x - 1)
    do
      result = result << 1
      if image.pixel(xx, yy) == Pixel.Lit then
        result = result | 1
      end if
    result
  end pixelValue
```
That way, it's not necessary to maintain `bitMask` at all.

sjrd: Also, if you use `end if`, you should use `end for`, since the `for` encloses the `if`.

src/day20.scala
julienrf: Fixed in 4a9a2e7

src/day20.scala
julienrf: Good point, thanks for the suggestion. Fixed in 4a9a2e7.

sjrd (approved):

PR #87, opened 2021-12-20
add Day 17 article

bishabosha:


docs/puzzles/day17.md
```diff
@@ -2,10 +2,410 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+
+We are also told about the projectile motion of the probe - it moves
+in discrete steps, which we will implement as follows:
+- on each step we create a new probe with a new position and a new velocity,
+- the `x` of the new position is the sum of `x` of the old position and `x`
+  of the old velocity,
+- the `y` of the new position is the sum of `y` of the old position and `y`
+  of the old velocity,
+- the `x` of the new velocity is given by subtracting the sign (`-1`, `0`
+  or `1`) of the `x` of the old velocity from itself (drag slows the
```
sjrd, line 50:
```suggestion
  or `1`) of the `x` of the old velocity from itself (drag slows down the
```

docs/puzzles/day17.md
```diff
@@ -2,10 +2,410 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+```scala
+case class Target(xs: Range, ys: Range)
+```
+
+### Reasoning about the Problem
+
+We are told to first identify the initial velocity of trajectories that will
+hit the target area exactly, meaning that after a given step, the probe will
+be exactly within the target area; and that we must disregard trajectories
+that "overshoot" i.e. pass through the entire target area in a single step.
```
sjrd, line 80:
```suggestion
that "overshoot", i.e., that pass through the entire target area in a single step.
```

docs/puzzles/day17.md
```diff
@@ -2,10 +2,410 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+
+We are also told to find the maximum height (`y` position) reached out of
+all valid trajectories.
+
+#### Simulating a trajectory
+
+For this problem we will simulate a probe moving along a trajectory, i.e.
```
sjrd, line 87:
```suggestion
For this problem we will simulate a probe moving along a trajectory, i.e.,
```

docs/puzzles/day17.md
```diff
@@ -2,10 +2,410 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+They are also informed by the fact that the probe will eventually always
+have negative velocity (due to gravity).
+:::
+
+##### Running the Simulation
+
+We can use our two conditions to now simulate the trajectory of a probe:
+- We begin with an initial `probe`, and an initial maximum height `maxY` of
+  `0`,
+- We will then iterate these values - on each iteration, we will apply
```
sjrd, line 138:
```suggestion
- We then iterate these values - on each iteration, we apply
```
Avoid the future tense here. There are more "will"'s down below, which should also be removed.

docs/puzzles/day17.md
```diff
@@ -2,10 +2,410 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+  `step` to the probe, and replace `maxY` by the maximum of `maxY` and
+  the `y` position of the current probe,
+- we will ignore any iteration steps where the `probe` does not collide
```
sjrd, line 141:
```suggestion
- we will ignore any iteration step where the `probe` does not collide
```

docs/puzzles/day17.md
```diff
@@ -2,10 +2,410 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+  with, or go beyond the target, i.e. the probe is still on a valid
```
sjrd, line 142:
```suggestion
  with, or go beyond the target, i.e., the probe is still on a valid
```

docs/puzzles/day17.md
```diff
@@ -2,10 +2,410 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+only get closer to `0`, so we will not need to consider
+negative `x` velocities.
+
+Second, we know that the problem requires us to find the highest positive
+height reached by the probe, and that the probe's `y` velocity can only
+fall once in motion. So we will not need to consider negative `y`
+velocities.
+
+That gives us the lower bounds for `x` and `y` velocities, what about
+the upper bound for the velocities?
```
sjrd, line 194:
```suggestion
the upper bounds for the velocities?
```

docs/puzzles/day17.md
```diff
@@ -2,10 +2,410 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+
+We know that a trajectory is invalid if it goes beyond the target in a
+single step, so the largest single step that the probe can move is the
+distance of the furthest corner of the target. For our problem that is the
+lower right corner, so our upper bounds will be the position of that corner.
```
sjrd, line 199: This explanation lacks the critical detail that, for *positive* initial y velocities, eventually there will be a point where the position is on the x axis (y position is 0) with a y velocity of `-initialYVelocity - 1`. This is why we can use `-bottomRightY` as upper bound for the y velocity.

docs/puzzles/day17.md
```diff
@@ -2,10 +2,410 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+From this input we will extract two `Range` values by pattern matching.
+
+Values of type `PartialFunction[A, B]` can be used as extractors in
+pattern matching, and as we are parsing strings, let's make a type alias
+`Parser[A]` to communicate our intent:
+```scala
+type Parser[A] = PartialFunction[String, A]
+```
+
+First, lets make a parser for `Int` values. We use a regex to check
```
sjrd, line 238:
```suggestion
First, let us make a parser for `Int` values. We use a regex to check
```

docs/puzzles/day17.md
bishabosha: I have now addressed this comment

sjrd (approved):

docs/puzzles/day17.md
```diff
@@ -2,10 +2,436 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+  Probe(
+    Position(x = pos.x + vel.x, y = pos.y + vel.y),
+    Velocity(x = vel.x - vel.x.sign, y = vel.y - 1)
+  )
+```
+
+#### A Target Area
+
+Next we are told that a successful launch trajectory causes the probe
+to be within in a target area after at least one of its steps. The area is
```
julienrf, line 68: Is this really correct to say “within in”?

julienrf (approved):

docs/puzzles/day17.md
sjrd: Ah no, it should be "within a target", indeed. ;)

bishabosha: good spot, I fixed it

julienrf:
You need to also update docusaurus.config.js to add “Day 17” to the dropdown in the navigation menu.

PR #90, opened 2021-12-21
Solution for day 21

sjrd:


src/day21.scala
```diff
@@ -0,0 +1,96 @@
...
+    for
+      die1 <- List(1, 2, 3)
+      die2 <- List(1, 2, 3)
+      die3 <- List(1, 2, 3)
+    yield
+      die1 + die2 + die3
+  possibleRolls.groupMapReduce(identity)(_ => 1L)(_ + _).toList
+
+def playWithDiracDie(players: Players, player1Turn: Boolean, wins: Wins, inHowManyUniverses: Long): Unit =
+  for (diesValue, count) <- dieCombinations yield
```
julienrf, line 83: Why bother with `yield`ing something in the absence of comonad?

sjrd: Oops, leftover from when I thought the functional way was going to be easier.

src/day21.scala
```diff
@@ -0,0 +1,96 @@
...
+    players(1).score
+  else
+    val newPlayer = Player(newCell, newScore)
+    playWithDeterministicDie((players(1), newPlayer), die)
+
+final class Wins(var player1Wins: Long, var player2Wins: Long)
+
+def part2(input: String): Long =
+  val players = parseInput(input)
+  val wins = new Wins(0L, 0L)
```
julienrf, line 67: Not really important, but `new` is not necessary.

julienrf (approved):

src/day21.scala
sjrd: I like using `new` explicitly when I create stateful instances.

PR #91, opened 2021-12-21
Article for day 21.

sjrd:


docs/puzzles/day21.md
```diff
@@ -1,11 +1,312 @@
...
+  val lines = input.split("\n")
+  (parsePlayer(lines(0)), parsePlayer(lines(1)))
+
+def parsePlayer(line: String): Player =
+  line match
+    case s"Player $num starting position: $cell" =>
+      Player(cell.toInt - 1, 0L)
+```
+
+The only thing worth nothing is that we use numbers 0 to 9 for the cells (the "spaces") instead of 1 to 10.
```
julienrf, line 31:
```suggestion
The only thing worth noting is that we use numbers 0 to 9 for the cells (the "spaces") instead of 1 to 10.
```

docs/puzzles/day21.md
```diff
@@ -1,11 +1,312 @@
...
+The only reason is that it simplifies computations for wrapping around.
+
+## The deterministic die
+
+For the first play with the deterministic die, we have to return the score of the losing player, while keeping tabs on how many times the die was rolled.
```
julienrf, line 36:
```suggestion
For the first part with the deterministic die, we have to return the score of the losing player, while keeping tabs on how many times the die was rolled.
```
?

docs/puzzles/day21.md
```diff
@@ -1,11 +1,312 @@
...
+We model the deterministic die as an instance of a class `DeterministicDie`:
+
+```scala
+final class DeterministicDie {
+  var throwCount: Int = 0
```
julienrf, line 41: realizing now that the “setter” should probably be private, and only the “getter” should be public. But you might say that it’d be overengineered.

docs/puzzles/day21.md
```diff
@@ -1,11 +1,312 @@
...
+        wins.player2Wins += 1L
+    else
+      val newPlayer = Player(newCell, newScore)
+      playWithDiracDie((players(1), newPlayer), !player1Turn, wins)
+  end for
+```
+
+For every possible outcome, we compute the new cell and score.
+If the game is over, we add 1 to the number of ties that the current player wins.
+Otherwise, we recursive for the next roll.
```
julienrf, line 193: Is it correct to say “we recursive for”?

julienrf (approved):
Interesting! (and brilliant, of course)

docs/puzzles/day21.md
```diff
@@ -1,11 +1,312 @@
...
+# Day 21: Dirac Dice
+by @sjrd
 
 ## Puzzle description
 
 https://adventofcode.com/2021/day/21
 
+## Modeling and parsing the input
+
+There are no more surprises here, so let us directly see the code:
```
adpi2, line 12: It's okay not to give much explanations but we should not pretend it is obvious for everyone. Someone may start here without knowing about previous solutions.

```suggestion
As we did for previous solution, we start by modelling the problem and parsing the input.
```

docs/puzzles/day21.md
sjrd: I thought about that but I self-concluded it would be overengineered, indeed.

docs/puzzles/day21.md
```diff
@@ -1,11 +1,312 @@
...
+
+```scala
+type Cell = Int // from 0 to 9, to simplify computations
+
+case class Player(cell: Cell, score: Long)
+
+type Players = (Player, Player)
+
+def parseInput(input: String): Players =
```
adpi2, line 21: I don't think `Players` is better than  `(Player, Player)` because it hides that there can only be 2 players.

```suggestion
def parseInput(input: String): (Player, Player) =
```

docs/puzzles/day21.md
sjrd: Nope. It's "we recurse", indeed. ;)

docs/puzzles/day21.md
adpi2, line 31: It worth something :)

```suggestion
The only thing worth noting is that we use numbers 0 to 9 for the cells (the "spaces") instead of 1 to 10.
```

sjrd:
Thanks for the review. Fixed in https://github.com/scalacenter/scala-advent-of-code/pull/92.

docs/puzzles/day21.md
adpi2: sorry I should have reviewed that earlier. It is okay if you leave it this way.

docs/puzzles/day21.md
```diff
@@ -1,11 +1,312 @@
...
+In the tail-recursive call, we swap out the two players, so that the next player to play always comes first.
+
+Other than that, the function is a direct translation from the rules of the game in the puzzle description:
+
+1. Throw the die 3 times, and add up the values
+2. Advance the player, wrapping around every 10 cells
+3. Compute the new score
+4. Detect the winning condition and return the loser's score, or continue with the next player
+
+If you are not familiar with modular arithmetics (anymore?), it may not be obvious that the computation `(player.cell + diesValue) % 10` is correct.
```
adpi2, line 81: The `(anymore ?)` kind of assumes that most people learned and forgot about modular arithmetics. But we should not assume any background in mathematics from our readers.

```suggestion
If you are not familiar with modular arithmetics, it may not be obvious that the computation `(player.cell + diesValue) % 10` is correct.
```

docs/puzzles/day21.md
```diff
@@ -1,11 +1,312 @@
...
+A more clearly correct computation would be:
```
adpi2, line 82: What does `clearly correct` means? It is correct or it is not?

```suggestion
Another correct computation would be:
```

docs/puzzles/day21.md
```diff
@@ -1,11 +1,312 @@
...
+
+```scala
+val diesValueMod10 = diesValue % 10
+val newCell = (player.cell + diesValueMod10) % 10
+```
+
+Indeed, for every group of 10 in `diesValue`, the player will actually not move, so it moves only by `diesValue % 10`.
+The other `% 10` wraps around the circular board.
+
+Modular arithmetics tell us that `(player.cell + (diesValue % 10)) % 10` is in fact equivalent to `(player.cell + diesValue) % 10`.
```
adpi2, line 92: Nice explanation!

docs/puzzles/day21.md
adpi2, line 193: I think the verb is missing in this sentence?

adpi2 (approved):
Thanks a lot Seb!

docs/puzzles/day21.md
sjrd: That was the point of the question mark, to say that it *may* be "anymore" for some people, and not at all for other people. But I could also remove it altogether.

docs/puzzles/day21.md
sjrd: The intent was "A more-clearly correct", not "A more clearly-correct". Perhaps "Another computation, which is perhaps more straightforward, would be:" ?

PR #96, opened 2021-12-23
Add day 23

adpi2:


src/day23.scala
```diff
@@ -0,0 +1,151 @@
...
+  
+  private def nextStops(amphipod: Amphipod, from: Position): Seq[Position] =
+    val destination = getDestination(amphipod)
+    from match
+      case Position(x, y) if x == amphipod.destinationX =>
+        destination match
+          case Some(destination) => Seq.empty
+          case None => hallwayStops
+      case Position(_, 1) => destination.toSeq
+      case _ => hallwayStops ++ destination
```
sjrd, line 85: Easy win here:
```suggestion
      case _ => hallwayStops
```
Any solution that involves going from one room directly to the destination is subsumed by first making a stop in the hallway between the rooms, then moving into the destination. So by exploring both going in the hallway *and* directly go to the destination, you're duplicating work.

src/day23.scala
```diff
@@ -0,0 +1,151 @@
...
+
+  private def getPath(start: Position, stop: Position): Seq[Position] =
```
sjrd, line 87: Another easy win: precompute all possible paths and store them in a `Map[(Position, Position), Seq[Position]]`.

src/day23.scala
```diff
@@ -0,0 +1,151 @@
...
+  yield (nextSituation, currentEnergy + energy)
+  
+  nextSituations.find((situation, energy) => situation.isFinal) match
+    case Some((_, energy)) => energy
+    case None =>
+      val allSituations = nextSituations.foldLeft(currentSituations) { case (acc, (situation, energy)) => 
+        val minEnergy = math.min(acc.getOrElse(situation, Int.MaxValue), energy)
+        acc.updated(situation, minEnergy)
+      }
+      if allSituations != currentSituations then
```
sjrd, line 124: This is probably quite expensive (it's an equality test between unordered collections). Isn't there any constant-time way you can detect whether you should go to `Amphipod.A`? Couldn't you do that always when `movingAmphipod == D`, for example?

src/day23.scala
```diff
@@ -0,0 +1,151 @@
...
+
+def readInput(): String =
+  Using.resource(Source.fromFile("input/day23"))(_.mkString)
+
+type Energy = Int
+
+enum Amphipod:
+  case A, B, C, D
+
+  def energy: Energy = this match
```
sjrd, line 26: Easy win: make all of these `val` parameters of `enum Amphipod`, so that you don't recompute them all the time.

src/day23.scala
```diff
@@ -0,0 +1,151 @@
...
+    val x = amphipod.destinationX
+    @tailrec
+    def tailApply(y: Int): Option[Position] =
+      positionToAmphipod.get(Position(x, y)) match
+        case Some(other) if other == amphipod => if y == 2 then None else tailApply(y - 1)
+        case Some(_) => None
+        case None => Some(Position(x, y))
+    tailApply(roomSize + 1)
+  
+  private def nextStops(amphipod: Amphipod, from: Position): Seq[Position] =
```
sjrd, line 77: I can't tell whether this method tries to avoid moving an amphipod that is in a room, not first in line. Such amphipods are blocked no matter where they try to go. So if it is not currently the case, an easy win is to return `Seq.empty` here for those.

sjrd (commented):
Some easy wins in terms of performance.

sjrd:
More complicated wins:

* Distinguish individual amphipods as separate instances that store their positions, rather than fixed A, B, C, D and having to remember a *set* of positions for each.
* Model the hallway and the rooms separately in the Situation. The hallway is a single `Seq` indexed by `x` coordinates. The rooms are each a `List[Amphipod]`, ordered from outermost (the one closest to the room) to innermost; remember whether the room is in "moving out" or "moving in" state (whether some pod still needs to get out of there, or all of them are of the right type). Only attempt to move the one that's at the `head` of the list.
* In Paths, independently store their physical size (which is important for the energy) from the list of valid hallway positions. That's less positions to check for `isEmpty`.

sjrd:
If that's not enough, it's interesting to look at what you did in `findCheapestSolution`:
```scala
      val allSituations = nextSituations.foldLeft(currentSituations) { case (acc, (situation, energy)) => 
        val minEnergy = math.min(acc.getOrElse(situation, Int.MaxValue), energy)
        acc.updated(situation, minEnergy)
      }
```
You are sort-of, but not quite, computing a shortest weighted path algorithm in a virtual graph where Situations are the nodes, and `moves` are the edges (the energy cost of each move being the weight of the edges).

You can apply a Dijkstra algorithm on that virtual graph. You have an ounce of it already in your way of checking whether you can improve the `energy` required to get to a certain situation. You can take it all the way with a `PriorityQueue` of the next situations you can encounter, and always treat the one with minimal energy next. You can stop the Dijkstra algorithm when you find a `isFinal` situation.

If *that* is *still* not good enough, well, we only care about the shortest path in that virtual graph between a single pair of nodes (Situations): the given initial situation, and the final situation. That means we an apply an A* algorithm in the virtual graph, instead of Dijkstra. For A* we need a *heuristic* that never over-estimates the "distance" (energy in our case) necessary to get from one situation to another. Such a heuristic can be: take the energy required to move every amphipod to its final destination, regardless of whether it will be blocked by other pods or not.

---

OK, I probably got carried away, so I'm going to stop here 😅

adpi2:
Thanks @sjrd the Djikstra algorithm makes it faster and the resulting code looks more readable.

src/day23.scala
```diff
@@ -0,0 +1,158 @@
...
+  /**
+   * Get the destination position of an amphipod in the current situation
+   * It returns `None` if the destination room is not free.
+   * It returns `Some(position)` if the destination room is free where `position` is the lowest
+   * empty position of the room.
+   * It returns `None` if the destination room is full.
+   */
+  private def getDestination(amphipod: Amphipod): Option[Position] =
+    if isDestinationFree(amphipod) then
+      (roomSize + 1).to(2, step = - 1)
```
sjrd, line 96:
```suggestion
      (roomSize + 1).to(2, step = -1)
```
I've never seen a coding style where we put a space between a unary operator and its argument.

src/day23.scala
```diff
@@ -0,0 +1,158 @@
...
+        (line, y) <- input.linesIterator.zipWithIndex
+        (char, x) <- line.zipWithIndex
+        amphipod <- Amphipod.tryParse(char)
+      yield Position(x, y) -> amphipod
+    Situation(positions.toMap, roomSize)
+
+class DjikstraSolver(initialSituation: Situation):
+  private val bestSituations = mutable.Map(initialSituation -> 0)
+  private val situationsToExplore =
+    mutable.PriorityQueue((initialSituation, 0))(Ordering.by((_, energy) => - energy))
```
sjrd, line 132:
```suggestion
    mutable.PriorityQueue((initialSituation, 0))(Ordering.by((_, energy) => -energy))
```

src/day23.scala
sjrd: I'm still a bit confused by this method. I wonder if the separate `getDestination` method obscures a bit what this method is doing. Since `getDestination` is called only from this method, have you considered inlining its content in the `case Position(_, 1)`?

sjrd (approved):

sjrd:
Looks nice overall :) Thanks :)

PR #98, opened 2021-12-29
Add day 22 article

bishabosha:
a holiday treat for you

### DIFF
--- a/docs/puzzles/day22.md
+++ b/docs/puzzles/day22.md
@@ -2,9 +2,345 @@ import Solver from "../../../../website/src/components/Solver.js"
 
 # Day 22: Reactor Reboot
 
+by [@bishabosha](https://twitter.com/bishabosha)
+
 ## Puzzle description
 
 https://adventofcode.com/2021/day/22
+
+## Final Problem
+
+### Modelling The Domain
+
+#### Basic Data Types
+
+We model the input as a series of steps, where each step has a
+command (either `"on"` or `"off"`), and
+a cuboid area. Each cuboid is modelled by its dimensions
+in each of the x, y, and z dimensions.
+A dimension is modelled by two numbers - its start and end points,
+i.e. the minimum and maximum values on a single line in that
+dimension. Here are the data types modelling the above:
+
+```scala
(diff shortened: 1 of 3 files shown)


docs/puzzles/day22.md
```diff
@@ -2,10 +2,346 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+Lastly we add another method to `Dimension` to get its size:
+```scala
+...
+  def size: Int = max - min + 1
+```
+
+#### Does one Cuboid `intersect` with Another?
+
+Now that we know how to get the intersection of a single
+dimension, we can extend this to the intersection of two
```
adpi2, line 86:
```suggestion
Now that we know how to get the intersection of two
dimensions, we can extend this to the intersection of two
```

docs/puzzles/day22.md
```diff
@@ -2,10 +2,346 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+      Cuboid(xs, ys, zs)
+```
+
+#### What is the `volume` of a Cuboid?
+
+With the `size` of each dimension determined, we can add a method
+to `Cuboid` to determine its `volume`, it is computed as a
+`BigInt` to avoid numeric overflow:
+```scala
+...
```
adpi2, line 107:
```suggestion
```

docs/puzzles/day22.md
```diff
@@ -2,10 +2,346 @@ import Solver from "../../../../website/src/components/Solver.js"
...
+We check that a cuboid is in the initialisation
+sequence with the following:
+
+```scala
+def isInit(cuboid: Cuboid): Boolean =
+  Seq(cuboid.xs, cuboid.ys, cuboid.zs).forall(_.isSubset(-50 by 50))
+```
+
+The final code for part 1 is then to run the steps
+unly while they fit the initialisation sequence, and then
```
adpi2, line 296:
```suggestion
only while they fit the initialisation sequence, and then
```

adpi2 (approved):
That's a good one! Thanks!

PR #110, opened 2022-12-01
Update website 2022

bishabosha:
- dynamically generate all links in sidebar, dropdowns, and index page.
- use source generators to patch the package of 2021 solutions (so we can use 2022 solutions if necessary)
- move all pages to 2021 links


solver/src/main/scala/adventofcode/Solver.scala
```diff
@@ -6,57 +6,71 @@ import scala.util.{Try, Failure, Success}
...
     for
-      solution <- solutions.get(puzzleId)
+      solutionOfYear <- solutions.get(year)
+      solution <- solutionOfYear.get(puzzleId)
       div <- Option(document.getElementById(puzzleId))
     do render(div, solverElement(solution))
 
+  @JSExportTopLevel("default")
+  def solver(puzzleId: String): Unit =
+    doSolve(puzzleId, "2021")
```
adpi2, line 68: Do we still need this one?

adpi2 (approved):
So cool! Thanks!

PR #158, opened 2022-12-07
Link to my day 7 solution

Jannyboy11:
Using mutable state this time since I'm not an immutable state-purist.
I think it's doable with immutable state only, but that requires that the parent's childDirectories map is updated once traversing back to the parent using `cd ..` and I wasn't sure whether the terminal output always includes that for every directory (it needs to cd back into the root directory for that to work). Also my solution accounts for `cd /` occuring multiple times in the input.


bishabosha (approved):
Thanks again!

PR #173, opened 2022-12-09
Add an article to day07

Decel:


docs/2022/puzzles/day07.md
```diff
@@ -1,14 +1,149 @@
...
+```
+
+##Final Solution
+
+```Scala
+import scala.annotation.tailrec
+import scala.collection.mutable
+import TerminalOutput.*
+import Command.*
+def input (str: String) = str.linesIterator.map {
```
bishabosha, line 92: please could you add vertical spacing (1 line) between the various definitions

docs/2022/puzzles/day07.md
```diff
@@ -1,14 +1,149 @@
...
+    case s"$$ cd $directory" => Cmd(ChangeDirectory(directory))
+    case s"$$ ls" => Cmd(ListFiles)
+    case s"dir $directory" => Directory(directory)
+    case s"$size $file" => File(size.toInt, file)
+}.toList
+enum Command:
+    case ChangeDirectory(directory: String)
+    case ListFiles
+enum TerminalOutput:
+    case Cmd(cmd: Command)
```
bishabosha, line 102: please could you indent across the file with 2 spaces.

docs/2022/puzzles/day07.md
```diff
@@ -1,14 +1,149 @@
...
+                         val parent: DirectoryStructure | Null)
+```
+We have to come up with a way to calculate directory size -- we can use [`sum`](https://www.scala-lang.org/files/archive/api/current/scala/collection/immutable/List.html#sum[B%3E:A](implicitnum:scala.math.Numeric[B]):B) for the size of all files in directory and define size of all of the following subdirectories recursively, which will take care of problem:
+
+```Scala
+def directorySize(dir: DirectoryStructure): Int =
+    dir.files.values.sum + dir.subDirectories.values.map(directorySize).sum
+```
+
+Now we need to create a function, to transfer input in directory form. For that we can use [`match`](https://docs.scala-lang.org/tour/pattern-matching.html) and separate input, -- for that we can use cases and recursion will do the rest for us:
```
bishabosha, line 49:
```suggestion
Now we need to create a function to build the directory structure from the input. For that we can use [`match`](https://docs.scala-lang.org/tour/pattern-matching.html) and separate input, -- for that we can use cases and recursion will do the rest for us:
```

docs/2022/puzzles/day07.md
```diff
@@ -1,14 +1,149 @@
...
+    case File(size, name) :: t =>
+        currentDir.files.put(name, size)
+        buildState(t, currentDir, rootDir)
+    case Directory(name) :: t =>
+        currentDir.subDirectories.put(name, DirectoryStructure(name, mutable.Map.empty, mutable.Map.empty, currentDir))
+        buildState(t, currentDir, rootDir)
+    case Nil => ()
+```
+
+And now, we need to assemble the program, in part one, we will just search for all directories with size smaller `100000`, and calculate the sum of their sizes.
```
bishabosha, line 66:
```suggestion
And now, we need to assemble the program, in part one, we will search for all directories with size smaller `100000`, and calculate the sum of their sizes. 
```

bishabosha: we should follow the inclusive language guide: https://docs.scala-lang.org/contribute/inclusive-language-guide.html#the-words-easy-simple-quick-and-trivial

docs/2022/puzzles/day07.md
```diff
@@ -1,14 +1,149 @@
...
 # Day 7: No Space Left On Device
+code by [Jan Boerman](https://twitter.com/JanBoerman95)
 
 ## Puzzle description
 
 https://adventofcode.com/2022/day/7
 
+## Solution
+
+First of all, we need to come up with a way to parse out input code:
```
bishabosha, line 12: Id swap the definition + explanation of the Command enum and the input parsing, I think it works better if you don't have forward references

docs/2022/puzzles/day07.md
```diff
@@ -1,14 +1,149 @@
...
+
+```Scala
+def input (str: String) = str.linesIterator.map {
```
bishabosha, line 15: this snippet should also include imports for Command and TerminalOutput

bishabosha (approved):
Thank you!

PR #177, opened 2022-12-10
Added solution to day 10

MewenCrespo:
Hope I followed the code convention well

### DIFF
--- a/2022/src/day10.scala
+++ b/2022/src/day10.scala
@@ -1,0 +1,48 @@
+package day10
+
+import locations.Directory.currentDir
+import inputs.Input.loadFileSync
+
+import Direction.*
+
+def INPUT(): String = loadFileSync(s"$currentDir/../input/day10")
+
+enum Command {
+  case NOOP
+  case ADDX(X: Int)
+}
+
+export Command.*
+
+def commandsIterator: Iterator[Command] = for (line <- INPUT.linesIterator) yield line.strip match {
+  case "noop" => NOOP
+  case s"addx $x" if x.toIntOption.isDefined => ADDX(x.toInt)
+  case _ => throw IllegalArgumentException(s"Invalid command '$line''")
+}
+
+val REGISTER_START_VALUE = 1
+
+def registerValuesIterator: Iterator[Int] = {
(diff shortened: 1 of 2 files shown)


2022/src/day10.scala
```diff
@@ -0,0 +1,48 @@
+package day10
+
+import locations.Directory.currentDir
+import inputs.Input.loadFileSync
+
+import Direction.*
+
+def INPUT(): String = loadFileSync(s"$currentDir/../input/day10")
```
bishabosha, line 8: please can you refactor the code so that there exists `def part1(input: String)` and `def part2(input: String)` that receive the input already read from the file, like all the other files in this directory - its necessary for the solver on the webpage

bishabosha (changes requested):
Thank you for submitting, just a comment about code conventions

2022/src/day10.scala
bishabosha: I also want to say its not typical in scala to use ALL_CAPS, please refactor to UpperCamelCase

2022/src/day10.scala
```diff
@@ -0,0 +1,48 @@
...
+
+export Command.*
+
+def commandsIterator: Iterator[Command] = for (line <- INPUT.linesIterator) yield line.strip match {
+  case "noop" => NOOP
+  case s"addx $x" if x.toIntOption.isDefined => ADDX(x.toInt)
+  case _ => throw IllegalArgumentException(s"Invalid command '$line''")
+}
+
+val REGISTER_START_VALUE = 1
```
bishabosha, line 23: same here, please refactor to UpperCamelCase, e.g
```suggestion
val RegisterStartValue = 1
```

bishabosha, line 23: same here, please refactor across the file to UpperCamelCase, e.g
```suggestion
val RegisterStartValue = 1
```

MewenCrespo: Thank you, it should be the case now

2022/src/day10.scala
MewenCrespo: I switched to pure function, forwarding the input: String parameter to each function. The signatures are now, I think, as you desired

bishabosha (approved):
thanks!

PR #178, opened 2022-12-10
Added detailed solution to day 10

MewenCrespo:
Hopefully I did it right. Although I don't think the solver is ready yet.

### DIFF
--- a/docs/2022/puzzles/day10.md
+++ b/docs/2022/puzzles/day10.md
@@ -1,7 +1,7 @@
 import Solver from "../../../../../website/src/components/Solver.js"
 
 # Day 10: Cathode-Ray Tube
-code by [Mewen Crespo](https://github.com/MewenCrespo)
+code and article by [Mewen Crespo](https://github.com/MewenCrespo) (reviewed by [Jamie Thompson](https://twitter.com/bishabosha))
 
 ## Puzzle description
 
--- a/docs/2022/puzzles/day10.md
+++ b/docs/2022/puzzles/day10.md
@@ -6,6 +6,149 @@ import Solver from "../../../../../website/src/components/Solver.js"
 
 https://adventofcode.com/2022/day/10
 
+## Solution
+
+Today's goal is to simulate the register's values over time. Once this is done, the rest falls in place rather quickly. From the puzzle description, we know there are two commands availaible: noop and addx. This can be implemented with a enum:
+
+```scala
+enum Command {
+    case NOOP
+    case ADDX(X: Int)
+  }
+```
+
+Now, we need to parse this commands from the string. This can be done using a simple for loop to match each line of the input. Assuming the input is a string stored in the variable INPUT then the following does the trick:
+
+```scala
+def commandsIterator: Iterator[Command] = for (line <- INPUT.linesIterator) yield line.strip match {
+    case "noop" => NOOP
+    case s"addx $x" if x.toIntOption.isDefined => ADDX(x.toInt)
+    case _ => throw IllegalArgumentException(s"Invalid command '$line''")
+  }
+```
+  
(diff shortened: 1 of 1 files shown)


docs/2022/puzzles/day10.md
```diff
@@ -6,6 +6,149 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+def CRTImage: String = CRTCharIterator.grouped(CRT_WIDTH).map(_.mkString).mkString("\n")
+
+@main def part2(): Unit = println(s"The CRT output is:\n$CRTImage")
+```
+
+### Run it in the browser
+
+#### Part 1
+
+<Solver puzzle="day09-part1" year="2022"/>
```
bishabosha, line 146:
```suggestion
<Solver puzzle="day10-part1" year="2022"/>
```

docs/2022/puzzles/day10.md
```diff
@@ -6,6 +6,149 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+
+#### Part 2
+
+<Solver puzzle="day09-part2" year="2022"/>
```
bishabosha, line 150:
```suggestion
<Solver puzzle="day10-part2" year="2022"/>
```

bishabosha (changes requested):
the solvers wont render in the page if the key doesn't exist, but still its best to put the `puzzle` key for the correct article

docs/2022/puzzles/day10.md
bishabosha: the puzzle key is significant because it is used to find the actual code to run the solution

MewenCrespo:
> the solvers wont render in the page if the key doesn't exist, but still its best to put the `puzzle` key for the correct article

Oh my bad, that's a poor copy paste sorry

docs/2022/puzzles/day10.md
```diff
@@ -1,11 +1,145 @@
...
+
+```scala
+def commandsIterator(input: String): Iterator[Command] = for (line <- input.linesIterator) yield line.strip match {
+  case "noop" => NOOP
+  case s"addx $x" if x.toIntOption.isDefined => ADDX(x.toInt)
+  case _ => throw IllegalArgumentException(s"Invalid command '$line''")
+}
+```
+  
+Here we used `linesIterator` to retreive the lines (it returns an `Iterator[String]`) and mapped every line using a for .. yield .. match comprehension. Note the use of `line.strip` to remove any white space character and the string interpolator `s` for an easy string parsing.
```
bishabosha, line 31:
```suggestion
Here we used `linesIterator` to retrieve the lines (it returns an `Iterator[String]`) and mapped every line using a for .. yield .. match comprehension. Note the use of `line.strip` to remove any white space character and the string interpolator `s` for an easy string parsing.
```

docs/2022/puzzles/day10.md
```diff
@@ -1,11 +1,145 @@
...
+
+:::Error checking
+Althought not necessary in this puzzle, it is a good practice to check the validity of the input. Here, we checked that the string matched with `$x` is a valid integer string before entering the second cas and returned an error if none of the first cases were matched.
+:::
+
+Now we are ready to compute the registers values. We choose to implement it as an `Iterator[Int]` which will return the register's value each cycle at a time. For this, we need to loop throught the commands. If the command is a noop, then the next cycle will have the same value. If the command is a addx x then the next cycle will be the same value and the cycle afterward will be `x` more. We immediatly notice the issue: the addx command generates two cycles whereas the noop command generates only one.
+
+To circumvent this issue, we choose to generate an `Iterator[List[Int]]` first which we'll flatten afterward. The first iterator is naturraly constructed using the scanLeft method. This yields the following code:
```
bishabosha, line 39:
```suggestion
To circumvent this issue, we choose to generate an `Iterator[List[Int]]` first which we'll flatten afterward. The first iterator is constructed using the scanLeft method. This yields the following code:
```

bishabosha (approved):
I'd merge but the spelling is important

PR #185, opened 2022-12-11
📝 include information about private leaderboard

twentylemon:
I posted earlier today asking if there was a private leaderboard including us nerds. I was eventually pointed to it, but it was a little more effort than it should need to be :)

This change simply notes the existence of such a private leaderboard. Paired with this, I strongly recommend a mod pin the message: https://discord.com/channels/632150470000902164/913451015246868530/1047707459990605865

I didn't want to include the leaderboard invite code directly on the page. Private leaderboards can only have 200 members, so I figure we should try to avoid bots from joining 🙃 . Those that pop into the discord are more likely to be humans.

### DIFF
--- a/docs/introduction.md
+++ b/docs/introduction.md
@@ -19,7 +19,8 @@ We will strive to only use the Scala standard library to solve the puzzles, so t
 
 In addition, if you have your own solution for a puzzle and would like to share it, you can add a link to a gist in the solution page of the puzzles.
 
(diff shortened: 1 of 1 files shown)


SethTisue:
> I strongly recommend a mod pin the message [Discord link]

done 👍

SethTisue (approved):

bishabosha (approved):
now the message is pinned I'll merge, thanks!

PR #201, opened 2022-12-15
Add Artem's solution to the Day 3

sierikov:


bishabosha (approved):
Thanks for contributing!

PR #218, opened 2022-12-20
Wireframe article for day 15

Sporarum:


docs/2022/puzzles/day15.md
```diff
@@ -6,6 +6,127 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+```scala
+case class Position(x: Int, y: Int)
+
+def parse(input: String): List[(Position, Position)] =
+  input.split("\n").toList.map{
+    case s"Sensor at x=$sx, y=$sy: closest beacon is at x=$bx, y=$by" =>
+      (Position(sx.toInt, sy.toInt), Position(bx.toInt, by.toInt))
+  }
+```
+
```
bishabosha, line 24: Please could you add here a brief overview of the algorithm you are about to implement before you introduce the following snippets

bishabosha (approved):

PR #227, opened 2022-12-24
📝 article for day 24

twentylemon:
I was pretty happy with my solution today (I like a good A*) and felt inspired enough to write about it. So, here we are!

Write up is based on my solution here: https://github.com/twentylemon/advent-of-code/blob/main/src/test/scala/org/lemon/advent/year2022/Day24Test.scala


bishabosha (approved):
Thanks so much for writing the article!

PR #229, opened 2022-12-26
Add solutions for days 12-25

cosminci:
Some files were missing the empty endl - added those as well.


adpi2 (approved):
Thanks for sharing your solutions!

PR #234, opened 2022-12-30
Update day01.md

ciuckc:
Added my own solution for the first day of AOC22


adpi2 (approved):
Nice solution, thanks!

PR #237, opened 2023-01-02
Update day02.md

ciuckc:
Ugly hardcoded solution, could ve done it better if I was splitting each line in two different tokens and matching each tokens to get the result


adpi2 (approved):
Thanks for sharing!

PR #238, opened 2023-01-02
Update day03.md

ciuckc:
solution for day3 learned a lot about intersections and transforming lists into tuples


adpi2 (approved):
Thanks!

PR #251, opened 2023-01-16
add day 16 code

Sporarum:
More article-friendly version of
https://gist.github.com/JavadocMD/ad657672282b2b547334f10bd15d3066


bishabosha (approved):

bishabosha:
I think for the article less nesting would be better, but thank you!

2022/src/day16.scala
```diff
@@ -0,0 +1,153 @@
...
+  // each step involves moving to a room with a useful valve and opening it
+  // we don't need to track each (empty) room in between
+  // we limit our options by only considering the still-closed valves
+  // and `valves` has already culled any room with a flow value of 0 -- no point in considering these rooms!
+
+  def recurse(path: List[Id], valvesLeft: Set[Id], timeLeft: Int, totalValue: Int): Int =
+    // recursively consider all plausible options
+    // we are finished when we no longer have time to reach another valve or all valves are open
+    valvesLeft
+      .flatMap{ id =>
```
bishabosha, line 105: I was gong to suggest for comprehension here, but that doesnt work for a single flatMap :/

bishabosha: unless you extract the result of the Option.when :)

bishabosha:
```scala
for
  id <- valvesLeft
  t <- locally {
    val current  = path.head
    val distance = map.routes(current)(id)
    timeLeft - distance - 1
  }
  if t > 0
yield
  val value = map.rooms(id).flow * t
  recurse(id :: path, valvesLeft - id, t, totalValue + value)
```
hmm not bad

SethTisue:
@bishabosha merge?

bishabosha (approved):

PR #252, opened 2023-09-15
Scala 3.3.1 (was 3.2.0)

SethTisue:
Review by @bishabosha?

I see that we don't have an overarching CI for all this. Shall we optimistically merge regardless?

PR #253, opened 2023-09-15
add solution for 2022 day 7

SethTisue:
depends on #252. I'll rebase this one after that is merged

### DIFF
--- a/2022/src/day07.scala
+++ b/2022/src/day07.scala
@@ -4,31 +4,35 @@ import locations.Directory.currentDir
 import inputs.Input.loadFileSync
 
 @main def part1: Unit =
-  val root: Directory = Directory("/")
-  run(loadInput(), List(root))
-  val solution =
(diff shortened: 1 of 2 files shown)


2022/src/day07.scala
```diff
@@ -0,0 +1,95 @@
...
+  val root: Directory = Directory("/")
+  run(loadInput(), List(root))
+  val solution =
+    allSubdirs(root)
+      .map(totalSize)
+      .filter(_ <= 100_000L)
+      .sum
+  println(s"The solution is $solution")
+
+@main def part2: Unit =
```
bishabosha, line 16: to integrate with the article which has a solver widget, would it be possible to delegate part1 and part2 to two extra functions that take a string parameter (the loaded input)

bishabosha (approved):
awesome, thanks!

PR #261, opened 2023-11-23
Make the day16 of 2022 implementation fastish on Scala.js.

sjrd:
(and really fast on the JVM)

This is the best I could come up with, in a reasonable amount of time, and without changing the *algorithm* itself.

### DIFF
--- a/2022/src/day16.scala
+++ b/2022/src/day16.scala
@@ -19,7 +19,7 @@ you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
    http://www.apache.org/licenses/LICENSE-2.0
-   
+
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -59,7 +59,7 @@ def computeRoutes(start: Id, neighbors: Id => List[Id]): Map[Id, Int] =
 
   case class State(frontier: List[(Id, Int)], scores: Map[Id, Int]):
 
-    private def getScore(id: Id): Int = scores.getOrElse(id, Int.MaxValue) 
+    private def getScore(id: Id): Int = scores.getOrElse(id, Int.MaxValue)
     private def setScore(id: Id, s: Int) = State((id, s + 1) :: frontier, scores + (id -> s))
 
     def dequeued: (Id, State) =
@@ -74,7 +74,7 @@ def computeRoutes(start: Id, neighbors: Id => List[Id]): Map[Id, Int] =
 
   object State:
     def initial(start: Id) = State(List((start, 0)), Map(start -> 0))
-  
+
   def recurse(state: State): State =
     if state.frontier.isEmpty then
       state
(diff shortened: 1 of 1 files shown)


2022/src/day16.scala
```diff
@@ -98,27 +98,36 @@ def bestPath(map: RoomsInfo, start: Id, valves: Set[Id], timeAllowed: Int): Int
...
+  def recurse(current: Id, timeLeft: Int, totalValue: Int): Int =
     // recursively consider all plausible options
     // we are finished when we no longer have time to reach another valve or all valves are open
-    valvesLeft
-      .flatMap{ id =>
-        val current  = path.head
-        val distance = map.routes(current)(id)
+    val routesOfCurrent = map.routes(current)
+    var bestValue = totalValue
+    for index <- 0 until valveCount do
```
bishabosha, line 111:
```suggestion
    for index <- valvesArray.indices do
```

bishabosha:
nice improvement!

2022/src/day16.scala
```diff
@@ -98,36 +98,39 @@ def bestPath(map: RoomsInfo, start: Id, valves: Set[Id], timeAllowed: Int): Int
...
-        if t > 0 then
-          // the value of choosing a particular valve (over the life of our simulation)
-          // is its flow rate multiplied by the time remaining after opening it
-          val value = roomsByIndices(index).flow * t
-          valvesIndicesLeft(index) = false
-          val recValue = recurse(id, t, totalValue + value)
-          valvesIndicesLeft(index) = true
-          if recValue > bestValue then
-            bestValue = recValue
+    for index <- valvesLookup.indices if valveIndexLeft(index) do
```
sjrd, line 117: That's not nearly as good. Even `indices` is not great. Putting the `if` inside the comprehension creates a dreadful `withFilter`.

2022/src/day16.scala
```diff
@@ -98,36 +98,39 @@ def bestPath(map: RoomsInfo, start: Id, valves: Set[Id], timeAllowed: Int): Int
...
-  def recurse(current: Id, timeLeft: Int, totalValue: Int): Int =
+  val valvesLookup = IArray.from(valves)
+  val valveCount = valvesLookup.size
+  val _activeValveIndices = Array.fill[Boolean](valveCount + 1)(true) // add an extra valve for the initial state
+  def valveIndexLeft(i: Int) = _activeValveIndices(i)
+  def withoutValve(i: Int)(f: => Int) =
+    try
+      _activeValveIndices(i) = false
+      f
+    finally _activeValveIndices(i) = true
```
sjrd, line 109: The `try..finally` is also going to cost something on JS. And it's useless since we don't catch exceptions here until the array gets thrown away.

bishabosha: alrighty

2022/src/day16.scala
bishabosha: isn't indices just sugar for `0 to foo.length`?

sjrd: There's just enough indirection that the Scala.js cannot stack-allocate the `Range` anymore, and therefore does not turn the `foreach` into a `while` loop.

2022/src/day16.scala
bishabosha: I have reverted the changes anyway #264

PR #300, opened 2023-12-02
Add link to my solution for day 2.

Jannyboy11:
A bit late to the party today. Solution uses a bunch of folds.


bishabosha (approved):
Thanks again!

PR #301, opened 2023-12-02
Update day02.md

bishabosha:
make parsing a bit more elegant


docs/2023/puzzles/day02.md
```diff
@@ -47,15 +47,14 @@ Let's fill in the `parse` function as follows:
 
 ```scala
 def parseColors(pair: String): Colors =
-  val Array(count0, color0) = pair.split(" ")
-  Colors(color = color0, count = count0.toInt)
+  val (s"$value $name") = pair: @unchecked
```
SethTisue, line 50: ((((((parens not needed)))))

PR #320, opened 2023-12-03
Update day03.md

iusildra:


iusildra:
Compact yes, but I don't think it's very performance-friendly 😂

PR #338, opened 2023-12-04
Add first draft of day04 solution

shardulc:
CC @bishabosha @sjrd


2023/src/day04.scala
```diff
@@ -0,0 +1,64 @@
...
+Card 6: 31 18 13 56 72 | 74 77 10 23 35 67 36 11
+""".strip()
+
+def countWinning(card: String): Int =
+  val numbers = card
+    .substring(card.indexOf(":") + 1)   // discard "Card X:"
+    .split(" ")
+    .filterNot(_.isEmpty())
+  val separatorIndex = numbers.indexOf("|")
+  val (_winningNumbers, _givenNumbers) = numbers.splitAt(separatorIndex)
```
sjrd, line 31: `indexOf` + `splitAt` can be done in one go with `span`:
```suggestion
  val (_winningNumbers, _givenNumbers) = numbers.span(_ != "|")
```

sjrd: Also, using leading `_`'s is not a common style, except of private fields that also have a public accessor. Consider using an explicit suffix like `winningNumberStrs` for example.

2023/src/day04.scala
```diff
@@ -0,0 +1,64 @@
...
+  val winningNumbers = _winningNumbers.map(Integer.parseInt).toSet
```
sjrd, line 32:
```suggestion
  val winningNumbers = _winningNumbers.map(_.toInt).toSet
```

2023/src/day04.scala
```diff
@@ -0,0 +1,64 @@
...
+  // drop the initial "|"
+  val givenNumbers = _givenNumbers.drop(1).map(Integer.parseInt).toSet
+  winningNumbers.intersect(givenNumbers).size
+
+def part1(input: String): String =
+  input.linesIterator
+    .map{ line =>
+      val winning = countWinning(line)
+      if winning > 0 then Math.pow(2, winning - 1).toLong else 0
```
sjrd, line 41: Try not to mix Longs and Ints. If you really want `Long`s, consider writing
```suggestion
      if winning > 0 then Math.pow(2, winning - 1).toLong else 0L
```
but if you actually want `Int`s, consider
```suggestion
      if winning > 0 then Math.pow(2, winning - 1).toInt else 0
```

PR #341, opened 2023-12-04
Solution day03 article

iusildra:
@bishabosha
I wrote the article as you suggested on my pr :slightly_smiling_face:

I tried to explain as much as possible, but if you think it needs some more explanations, I'll add them :grin:

### DIFF
--- a/docs/2023/puzzles/day03.md
+++ b/docs/2023/puzzles/day03.md
@@ -26,7 +26,7 @@ The solution models the input as a grid of numbers and symbols.
        4. Sum the resulting products
      - a symbol is adjacent to a number (and vice-versa) if that symbol is inside the number's bounding box on the grid at 1 unit away (see manhattan distance)
 
-### Framework
+### Global
 
 We want a convenient way to represent a coordinate to be able to compute whether one element is within the bounding box of another.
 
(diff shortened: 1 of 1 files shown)


docs/2023/puzzles/day03.md
```diff
@@ -11,187 +11,184 @@ https://adventofcode.com/2023/day/3
...
+    else true
+case class PartNumber(value: Int, start: Coord, end: Coord)
+case class Symbol(sym: String, pos: Coord):
+  def neighborOf(number: PartNumber) = pos.within(
+    Coord(number.start.x - 1, number.start.y - 1),
+    Coord(number.end.x + 1, number.end.y + 1)
+  )
+
+object IsInt:
+  def unapply(in: Match): Option[Int] = in.matched.toIntOption
```
bishabosha, line 155: your code is missing the `import scala.util.matching.Regex.Match`

bishabosha:
> With my solution it gives the same wrong result, and the output was the same as in `input/day03`

I'm still a bit confused what you mean - I copied your code into the repl and it gives the right output when using `input/day03`, so when is it wrong?

iusildra:
It's wrong when I run it from the website (after I updated the solution in the `solution` submodule)
image

iusildra:
Oh, well it seems that my copy/paste was the culprit :skull:

PR #356, opened 2023-12-04
Add day 4 solution by Will

wbillingsley:
There's actually 3 solutions for part 2 in my repo; 'star8b' is probably the best actual solution though. 'star8' was my original solution; 'star8c' was a gimmick from the Discord chat speculating on how to do it without any folds.


bishabosha (approved):
thanks again!

PR #361, opened 2023-12-05
add literate coding component

bishabosha:
Update day 4 article to use a new `<Literate>` react component that will group code blocks and paragraphs as follows:

Screenshot 2023-12-05 at 11 57 04
Screenshot 2023-12-05 at 11 57 14

~~~mdx

### String Manipulation

Here is code
<Literate>

```scala
def foo(a: String): Int =
```
we're going to do fancy stuff with this `String`.
```scala
  val length = s.size
  val hash = s.hashCode
```
Now something even more wild!

```scala
  length + hash
end foo
```

</Literate>

~~~

### DIFF
--- a/docs/2023/puzzles/day04.md
+++ b/docs/2023/puzzles/day04.md
@@ -1,4 +1,5 @@
 import Solver from "../../../../../website/src/components/Solver.js"
+import Literate from "../../../../../website/src/components/Literate.js"
 
 # Day 4: Scratchcards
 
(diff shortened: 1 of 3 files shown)


bishabosha:
@shardulc how does this look?

bishabosha:
seems like a classic case of "works on my machine" but not on the actually served page

shardulc:
Live version looks fine to me! I’d just suggest that the comment “Lastly, the standard library conveniently gives us an iterator over lines.” should instead be an interspersed comment like the ones above it, but not a big deal.

PR #368, opened 2023-12-05
Add g.berezin solution for day 5 with explanation

GrigoriiBerezin:


bishabosha:
Hello, thanks for taking the time to write the solution! however I'd ask you to consider the [schedule](https://github.com/scalacenter/scala-advent-of-code/discussions/348) first - @stewSquared is meant to be doing today

stewSquared:
@bishabosha @GrigoriiBerezin 
I was just now sitting down to write this up.
Seeing as how this is already written, I don't mind handing it over.
Perhaps I'll do some review and provide suggested edits.

bishabosha:
ok great, thank you both :)

stewSquared:
@GrigoriiBerezin Do you have your original code published on GitHub?

I'm finding it a little tricky to get code from your write-up compiling (endFrom is missing, uses some methods from cats, it's not always clear when a def is a method and what class it's supposed to be part of).

docs/2023/puzzles/day05.md
```diff
@@ -6,6 +6,151 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
 https://adventofcode.com/2023/day/5
 
+## Solution summary
+
+First and foremost, the data must be parsed from the file into the following classes:
+- ```final case class Resource(start: Long, `type`: DestinationEnum)``` the source resource in the first task
+- ```final case class ResourceWithRange(start: Long, end: Long, `type`: DestinationEnum)``` the source resource in the second task, in the first one you can use this same model but set `end = 0`
+- I used the following enumeration for storing the type of resource:
+```scala
+object DestinationEnum extends Enumeration {
```
stewSquared, line 16: I think it's worth using Scala 3 for blog posts. This particular example would be:
```scala
enum Destination:
  case Seed, Soil, Fertilizer, Water, Light, Temperature, Humidity, Location
```

docs/2023/puzzles/day05.md
```diff
@@ -6,6 +6,151 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+  type DestinationEnum = Value
+  val Seed, Soil, Fertilizer, Water, Light, Temperature, Humidity, Location = Value
+}
+```
+- ```final case class ResourceMap(from: DestinationEnum, to: DestinationEnum, properties: Seq[Property])``` and ```final case class Property(startTo: Long, startFrom: Long, range: Long)``` for storing data about converting from one resource type to another
```
stewSquared, line 21: Since property directly corresponds to the Advent of Code input, I'd name it morke like it is in the problem.

`startTo` is called "destination range start"
`startFrom` is called "source range start"
`range` is called "range length"

The `DestinationEnum`, `from` and `to` are called the "source category" and "destination category"

Do with that what you will.

bishabosha:
Just a heads up, we have a requirement that the code be written with a clear `def part1(input: String): Any` and `def part2(input: String): Any` entry points (the result can be more specialised)

GrigoriiBerezin:
> Hello, thanks for taking the time to write the solution! however I'd ask you to consider the [schedule](https://github.com/scalacenter/scala-advent-of-code/discussions/348) first - @stewSquared is meant to be doing today - hopefully we can agree together what should be done here
> 
> Edit: I should probably fix the website template so that people look first there

Sorry, didn't see that you have schedule for that

GrigoriiBerezin:
I've solved task using scala2, but I will reformat solution explanation to scala3 with your edits today, if you don't mind

GrigoriiBerezin:
> Just a heads up, we have a requirement that the code be written with a clear `def part1(input: String): Any` and `def part2(input: String): Any` entry points (the result can be more specialised)
> 
> Also i believe the mapping is always sequential so mostly the resource "kind" is irrelevant (please correct me if some input had a non sequential chain)

In case of changing order of maps, should I left it as it is? User input may be different due to no limitation about it in task

bishabosha:
I refactored a bit the code, it's still the same solution but reuses `calculate` from part 2 for both solutions. I also described the functionality of the code

GrigoriiBerezin:
> Thanks again for refactoring!
> 
> 
> 
> I refactored a bit the code, it's still the same solution but I tried to make it use up fewer columns (avoiding scrolling), also it reuses `calculate` from part 2 for both solutions. I also described the functionality of the code more.

Thank you

PR #377, opened 2023-12-06
Added day06 solution article

spamegg1:
- Please feel free to modify, if there are issues with formatting, or if the code (especially parsing) could be better.
- I'm unfamiliar with the format the website uses, and the "Run in the browser" boxes. I tried to follow Day 04 solution format as much as possible.
- You can see it working [here](https://scastie.scala-lang.org/spamegg1/j61U1JqcRxKjpgkGWHBqkQ/4)


bishabosha (approved):
Thank you for taking the time to write this up, it looks great!

PR #393, opened 2023-12-06
fixing small mistake in day06 solution article

spamegg1:
Looks like I forgot the line `val disc = math.sqrt(t * t - 4 * d)` in `solve`.

My apologies!


bishabosha (approved):
thanks! it compiles locally

PR #437, opened 2023-12-10
day 9 article

SethTisue:
feedback/suggestions welcome

### DIFF
--- a/docs/2023/puzzles/day09.md
+++ b/docs/2023/puzzles/day09.md
@@ -69,6 +69,10 @@ Note that doing `xs.tail.lazyZip(xs)` instead of the more obvious
 `xs.lazyZip(xs.tail)` means we don't need to reverse the arguments
 before subtracting, allowing use of the `_ - _` shorthand.
 
+Note also that both `zip` and `lazyZip` discard any extra values,
+so it doesn't matter that one of the sequences being zipped is one
+shorter.
+
 It would be only slightly more awkward to instead use `sliding(2)` to
 iterate over successive pairs, as follows:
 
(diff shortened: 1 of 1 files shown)


sjrd (approved):
Looks great.

I have no idea about the math background, though. If you're sure about that, we can merge.

docs/2023/puzzles/day09.md
```diff
@@ -2,10 +2,128 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+_next_ row would be all zeros.)
+
+The use of
+[`lazyZip`](https://www.scala-lang.org/api/current/scala/collection/Seq.html#lazyZip-fffffd84)
+instead of regular `zip` is also a small optimization that avoids
+constructing an intermediate collection.
+
+Note that doing `xs.tail.lazyZip(xs)` instead of the more obvious
+`xs.lazyZip(xs.tail)` means we don't need to reverse the arguments
+before subtracting, allowing use of the `_ - _` shorthand.
```
sjrd, line 70: Perhaps one thing here: it might be worth reminding that `zip`/`lazyZip` discard extra elements in the longer sequence. More naively we would expect `xs.tail.lazyZip(xs.init)`, and it is because of that special behavior that the `init` can be avoided.

SethTisue: added

SethTisue:
> I have no idea about the math background, though. If you're sure about that, we can merge.

I specifically asked about that in the Discord room, since I wasn't absolutely sure either. I think we can assume it's okay, given that extra review.

PR #441, opened 2023-12-10
Update day10.md

EugeneFlesselle:


bishabosha (approved):
looks great, I just added a few formatting changes

docs/2023/puzzles/day10.md
```diff
@@ -132,8 +138,12 @@ def findLoop(grid: Seq[String]): Seq[(Int, Int)] =
     val startI = grid.indexWhere(_.contains('S'))
     (startI, grid(startI).indexOf('S'))
 
-  /** List of connected points starting from 'S' (p0, p1) :: (p1, p2) :: (p2, p3) :: ... */
-  val loop = LazyList.iterate((start, connected(grid)(start).head)): (prev, curr) =>
+  val initial = (start, connected(grid)(start).head)
+
+  /** List of connected points starting from 'S'
+   * e.g. `(y0, x0) :: (y1, x1) :: (y2, x2) :: ...`
```
EugeneFlesselle, line 144: If we want to show both coordinates then it should be something like this:
`((y0, x0), (y1, x1)) :: ((y1, x1), (y2, x2)) :: ((y2, x2), (y3, x3)) :: ...`

EugeneFlesselle: Not supper important, but we could probably still fix it
@bishabosha

bishabosha: Oh right it's connected points oops

PR #453, opened 2023-12-12
Add solution for day 11 on the website

natsukagami:


docs/2023/puzzles/day11.md
```diff
@@ -6,6 +6,301 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+        val distanceToPointsHere = head * newTotalDist
+        loop(newCountsSum, newTotalDist, accum + distanceToPointsHere)(tail)
+```
+
+... and this is the solution presented in [our repo](https://github.com/scalacenter/scala-advent-of-code)!
+
+### Where to expand on the problem?
+
+Here are some ideas that I think would be interesting to look into:
+- Bigger inputs: what if the space is and incredibly large grid, but the number of `#`s are sparse (for example, about 100000 points in a 10^9-sized grid)?
```
SethTisue, line 299:
```suggestion
- Bigger inputs: what if the space is an incredibly large grid, but the number of `#`s are sparse (for example, about 100000 points in a 10^9-sized grid)?
```

bishabosha (approved):
phenomenal

PR #466, opened 2023-12-13
Add blog post for day 13

adpi2:


bishabosha (approved):
very nice

SethTisue:
just a nitpick, but `leftPart.reverse.zip(rightPart).forall(_ == _)` can be shortened to just `leftPart.reverse == rightPart`, can it?

SethTisue:
though that doesn't lead as naturally into the `l1.zip(l2).count(_ != _)` of part 2

SethTisue:
anyway, very nice solution, substantially more elegant than mine

it didn't occur to me to `type Tile = '.' | '#'`, probably because I've been doing Scala 2 for too long :-P

adpi2:
> just a nitpick, but leftPart.reverse.zip(rightPart).forall(_ == _) can be shortened to just leftPart.reverse == rightPart, can it?

It's not exactly the same because `leftPart` and `rightPart` can be of different size and we need to trim the bigger one.

SethTisue:
Ah, I see! In my code I trim separately first.

PR #471, opened 2023-12-14
Add day  14 blog

anatoliykmetyuk:


docs/2023/puzzles/day14.md
```diff
@@ -2,10 +2,357 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+    val array = Array.fill(extentX)(Array.fill(extentY)('.'))
+    for
+      (line, y) <- lines.zipWithIndex
+      (char, x) <- line.zipWithIndex
+    do array(x)(y) = char
+    dataN = array
+
+  override def toString =
+    val sb = new StringBuilder
+    for y <- (0 until extentY).toIterator do
```
bishabosha, line 63: doesn't compile

docs/2023/puzzles/day14.md
```diff
@@ -2,10 +2,357 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+  def setState(newState: String): Unit =
+    val lines = newState.split('\n').toList
+    val extentX = lines(0).length
+    val extentY = lines.length
+    val array = Array.fill(extentX)(Array.fill(extentY)('.'))
+    for
+      (line, y) <- lines.zipWithIndex
+      (char, x) <- line.zipWithIndex
+    do array(x)(y) = char
+    dataN = array
```
bishabosha, line 59:
```suggestion
  def setState(newState: String): Unit =
    dataN = newState.split('\n').map(_.toArray)
```
?

docs/2023/puzzles/day14.md
```diff
@@ -2,10 +2,357 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+
+  override def toString =
+    val sb = new StringBuilder
+    for y <- (0 until extentY).toIterator do
+      for x <- (0 until extentX).toIterator do
+        sb.append(this(x, y))
+      sb.append('\n')
+    sb.toString
```
bishabosha, line 67:
```suggestion
    dataN.map(_.mkString).mkString("\n")
```
?

docs/2023/puzzles/day14.md
```diff
@@ -2,10 +2,357 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+```scala
+def cycle(model: Model, times: Int): Unit =
+  val cache: collection.mutable.Map[String, String] = collection.mutable.Map.empty
+  var currentState = model.toString
+  for i <- 1 to times do
+    if i % 10_000_000 == 0 then
+      println(s"Cycle $i of ${times}; cache size: ${cache.size}")
+
+    if cache.keySet.contains(currentState) then
+      currentState = cache(currentState)
```
bishabosha, line 209: actually at the first time you get a cache hit you can compute the result at 1 billion in 1 step and return

bishabosha: see https://github.com/bishabosha/advent-of-code-2023/blob/a612e0dc72943e95677ab924243677ebafc2b12e/2023-day14.scala#L93C1-L108C10

bishabosha:
btw this solution does the entire rotation and cycling in about 2 lines:
https://github.com/merlinorg/aoc2023/blob/main/src/main/scala/Day14.scala

docs/2023/puzzles/day14.md
anatoliykmetyuk: Strange, works fine on my machine. Are you using the Scala 3.3.1? What error are you getting?

docs/2023/puzzles/day14.md
anatoliykmetyuk: What my code does is ensuring the `array(x)(y)` syntax. That is, when addressing the array, columns get addressed first (`x`), and rows - second (`y`). This is more in line with the mathematical view of the problem as a coordinate system, where the convention is to address the horizontal axis first. The simpler `newState.split('\n').map(_.toArray)` would have rows - the vertical axis - addressed first.

Going into a bit of trouble getting the coordinates right allows to use `x` and `y` in code, as opposed to `row` and `col`, which helps with understanding.

docs/2023/puzzles/day14.md
anatoliykmetyuk: `toString` here returns the *rotated* representation of the platform (see `this(x, y)`), as specified by the `rotation` variable`. `dataN.map(_.mkString).mkString("\n")` is not equivalent, as it always returns the default rotation of the platform - facing North.

docs/2023/puzzles/day14.md
anatoliykmetyuk: Nice! I implemented the shortcut solution.

docs/2023/puzzles/day14.md
bishabosha: this snippet has no field called extendY or extentX

docs/2023/puzzles/day14.md
bishabosha: right, sorry I didn't notice the y and x swap

bishabosha: but you will get the same behaviour with a transpose I believe:
```scala
scala> "012\n345".split('\n').map(_.toArray).transpose
val res3: Array[Array[Char]] = Array(Array(0, 3), Array(1, 4), Array(2, 5))
```

anatoliykmetyuk: Right, transpose also worked. I updated the article.

docs/2023/puzzles/day14.md
```diff
@@ -2,10 +2,337 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+The approach from Part 1 doesn't work here: we now need to modify our model state. The reason we didn't have to do it in Part 1 is that our model evolves in only a single step, and we only need one metric from the end state, so we don't have to calculate the entire final state. In Part 2, however, the model evolves in many steps, and each successive step depends on the previous step. So, we need to calculate the entire final state.
+
+### The Model
+Actually, we already know how to do the Northern tilt - similarly to Part 1. We do not want to re-implement the tilts for the other directions - instead, we want to use the Northern tilt as a building block for the other tilts. So, we need to be able to rotate our model in all four directions, and then tilt it North.
+
+We can represent the model as a 2D array of chars:
+
+```scala
+class Model(state: String):
+  private var dataN: Array[Array[Char]] = Array.empty
```
bishabosha, line 47: maybe add a comment that the array is indexed as `dataN(x)(y)` for a given point (x, y)

anatoliykmetyuk: Done

docs/2023/puzzles/day14.md
```diff
@@ -2,10 +2,337 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+
+So, there's a possibility that the desired state will be achieved earlier than the billion cycles. Looks like a good case for a dynamic programming approach: we need to remember the states of the model we've seen before and what they look like after one cycle. And if the current state is in our cache, no need to compute the next state again.
+
+Furthermore, since the transitions between states are deterministic (a state A always leads to state B), the moment one state in our cache is followed by another, previously encountered, state from that cache, we can stop cycling and just calculate the final state from the cache.
+
+```scala
+import scala.util.boundary, boundary.break
+
+def cycle(model: Model, times: Int): Unit =
+  val chain = collection.mutable.ListBuffer.empty[String]
```
bishabosha, line 198: If the aim here is all about optimisation shouldn't we not have a linear lookup?

docs/2023/puzzles/day14.md
```diff
@@ -2,10 +2,337 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+  val model = Model(input)
+  rollUp(model)
+  val loading = totalLoading(model)
+  println(loading)
+
+def part2(input: String) =
+  val model = Model(input)
+  cycle(model, 1_000_000_000)
+  val loading = totalLoading(model)
+  println(loading)
```
bishabosha, line 333:
```suggestion
def part1(input: String): Int =
  val model = Model(input)
  rollUp(model)
  totalLoading(model)

def part2(input: String): Int =
  val model = Model(input)
  cycle(model, 1_000_000_000)
  totalLoading(model)
```

bishabosha (approved):
alright, looks nice!

PR #490, opened 2023-12-16
Article day16

iusildra:
This time i managed to do it :grin: 
The file is [here](https://github.com/iusildra/advent-of-code-2023-scala/blob/main/src/main/scala/16.worksheet.sc)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ target/
 .scala-build
 .scala/
 website/src/js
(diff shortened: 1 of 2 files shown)


bishabosha (approved):
Thanks! its a good read - you might want to check out this about [sparse arrays](https://www.geeksforgeeks.org/what-is-meant-by-sparse-array/)

PR #513, opened 2023-12-21
Add an Article for Day 17 of 2023

stewSquared:


stewSquared:
This has a little errata. See https://github.com/scalacenter/scala-advent-of-code/pull/514
I'll handle it when I have a sec

docs/2023/puzzles/day17.md
```diff
@@ -2,12 +2,275 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+
 ## Puzzle description
 
 https://adventofcode.com/2023/day/17
 
+## Solution Summary
+
+This is a classic search problem with an interesting restriction on state transformations.
+
+We will solve this using Djikstra's Algorithm to find a path through the grid, using the heat loss of each position as our node weights. However, the states in our priority queue will need to include more than just position and accumulated heat loss, since the streak of forward movements in a given direction affects which positions are accessible from a given state.
```
Philippus, line 15:
```suggestion
We will solve this using Dijkstra's Algorithm to find a path through the grid, using the heat loss of each position as our node weights. However, the states in our priority queue will need to include more than just position and accumulated heat loss, since the streak of forward movements in a given direction affects which positions are accessible from a given state.
```

docs/2023/puzzles/day17.md
```diff
@@ -2,12 +2,275 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+    State(pos.move(newDir), newDir, 1)
+
+  def turnRight: State =
+    val newDir = dir.turnRight
+    State(pos.move(newDir), newDir, 1)
+```
+
+Note that the streak resets to one when we turn right or turn left, since we also move the position forward in that new direction.
+
+### Djikstra's Algorithm
```
Philippus, line 99:
```suggestion
### Dijkstra's Algorithm
```

docs/2023/puzzles/day17.md
```diff
@@ -2,12 +2,275 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+
+Finally, let's lay the groundwork for an implementation of Djikstra's algorithm.
```
Philippus, line 101:
```suggestion
Finally, let's lay the groundwork for an implementation of Dijkstra's algorithm.
```

Philippus (changes requested):
A typo in Dijkstra's name. :)

bishabosha (approved):
Really nice, thank you! I did some refactoring so that the grid isn't stored as a top level global

bishabosha:
Throwing out ideas, we do have this requirement of part1 and part2 not using IO to load the input so its scala.js compatible but perhaps we can add an extra section to each page that covers how to load the input from file that then calls the part1, part2 methods? it could even be tabbed for scala.js/scala native/jvm

stewSquared:
@bishabosha Thanks! I like the change removing global vars with Using Grid. I'll use that pattern in future submissions (and day 23, which I'm writing up now)

re: not using IO: Yes! I think that's a great idea. Should we make it work with pastable input in 2024?

bishabosha:
> re: not using IO: Yes! I think that's a great idea. Should we make it work with pastable input in 2024?

That's what the solver widget is meant to be if I take the time to set them up

PR #545, opened 2024-01-10
Add 2023 day 23 solution and input

stewSquared:


2023/input/day23
```diff
@@ -0,0 +1,23 @@
+#.#####################
```
bishabosha, line 1: Apparently these inputs shouldn't be shared (look it up), so now we will no longer accept them :)

stewSquared: I know! These aren't my inputs. They're the example :)

bishabosha: well I think even the sample input is still protected

bishabosha:
Lovely solutions all around, thank you again!

PR #560, opened 2024-12-01
Add Day 1 solution by Quentin Bernet

Sporarum:
Probably won't write an article, but I added comments to make it clearer


adpi2 (approved):
Thank you!

PR #566, opened 2024-12-02
Solution article for 2024 day 02

spamegg1:
[Here](https://scastie.scala-lang.org/ey2v1yv5QwCBF7h0nqm5CA) is a convenient Scastie of the solution.

### DIFF
--- a/docs/2024/puzzles/day02.md
+++ b/docs/2024/puzzles/day02.md
@@ -22,7 +22,7 @@ Therefore parsing looks like this:
 ```scala
 case class Report(levels: Seq[Long])
 
-def parseLine(line: string): Report = Report(line.split(" ").map(_.toLong).toSeq)
+def parseLine(line: String): Report = Report(line.split(" ").map(_.toLong).toSeq)
 
 def parse(input: String): Seq[Report] = input
   .linesIterator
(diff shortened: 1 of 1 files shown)


SethTisue:
thank you!!

SethTisue:
hmm, should I not have merged this until tomorrow? 🤔   @bishabosha ?

bishabosha:
For spoilers? I don't think we had any policy before like that

bishabosha:
it would be cool if there was some react/docusaurus package for spoiler tags (combined with a date filter?)

SethTisue:
an argument for merging promptly is that then the page is available for community folks to add their links to

I think that probably outweighs any spoiler concerns

bishabosha:
the page is already there in the morning when the cron job bot creates it about few mins after the challenge goes live

bishabosha:
anyway I did find this interesting package https://github.com/abdullahtariq1171/react-spoiler, which I guess could put the whole article sandwich between spoiler tags

SethTisue:
> the page is already there in the morning when the cron job bot creates it about few mins after the challenge goes live

oh, okay. in that case, I will try to hold off merging these PRs until the following day. it also gives more time for review, anyway

PR #605, opened 2024-12-04
add day 4 article

bishabosha:
did some refinements along the way (iterator over recursion)

### DIFF
--- a/docs/2024/puzzles/day04.md
+++ b/docs/2024/puzzles/day04.md
@@ -2,9 +2,290 @@ import Solver from "../../../../../website/src/components/Solver.js"
 
 # Day 4: Ceres Search
 
+by [@bishabosha](https://github.com/bishabosha)
+
 ## Puzzle description
 
 https://adventofcode.com/2024/day/4
+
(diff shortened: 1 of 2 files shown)


docs/2024/puzzles/day04.md
```diff
@@ -2,10 +2,289 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+  scanner(x, y, dir, grid).take(target.length).corresponds(target)(_ == _)
+
+val scanXMAS = scanString("XMAS")
+```
+
+Now we have a way to find a string in a grid, we should iterate through each point in the grid, and count all the possible `"XMAS"` strings starting from that point:
+
+```scala
+def totalXMAS(grid: Grid): Int =
+  var total = 0
```
SethTisue, line 125: Seems a bit odd to resort to a `var` here? Not a blocker for merging

SethTisue (commented):
otherwise LGTM, but let's not merge until day 5 puzzle is out

KacperFKorban (approved):
Looks great!

PR #633, opened 2024-12-06
Day06 part 1

samuelchassot:
Here is the solution for today :)

### DIFF
--- a/docs/2024/puzzles/day06.md
+++ b/docs/2024/puzzles/day06.md
@@ -1,6 +1,10 @@
 import Solver from "../../../../../website/src/components/Solver.js"
 
 # Day 6: Guard Gallivant
+
+by [@samuelchassot](https://github.com/samuelchassot)
+
+
 # Day 06 - solution
 
 ## Part 1
--- a/docs/2024/puzzles/day06.md
+++ b/docs/2024/puzzles/day06.md
@@ -1,7 +1,6 @@
 import Solver from "../../../../../website/src/components/Solver.js"
 
 # Day 6: Guard Gallivant
-
 # Day 06 - solution
 
 ## Part 1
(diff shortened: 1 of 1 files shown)


docs/2024/puzzles/day06.md
```diff
@@ -1,8 +1,214 @@
...
+      )
+
+    val parList = newPossibleGuards.par
+    val forkJoinPool = new ForkJoinPool(8)
+    parList.tasksupport = ForkJoinTaskSupport(forkJoinPool)
+
+    parList.count(g => looping(g, startingPov))
+```
+
+As we are brute forcing, let's use the parallel collection, to use all the potential of our nice hardware, to check multiple lab dispostions at the same time. We also use a `ForkJoinPool` to control the number of threads used by the parallel collection. See how simple it is to use parallel collections in Scala, using the `.par` method.
```
SethTisue, line 208:
```suggestion
As we are brute forcing, let's use a parallel collection, to use all the potential of our nice hardware, to check multiple lab dispositions at the same time. We also use a `ForkJoinPool` to control the number of threads used by the parallel collection. See how simple it is to use parallel collections in Scala, using the `.par` method.
```

docs/2024/puzzles/day06.md
```diff
@@ -1,8 +1,214 @@
...
+      )
+
+    val newPossibleGuards= 
+      possibleObstaclesPositions.map(obstaclePos => 
+        val newLab = g.lab.replaceWith(obstaclePos._1, obstaclePos._2, '#')
+        Guard(newLab)
+      )
+
+    val parList = newPossibleGuards.par
+    val forkJoinPool = new ForkJoinPool(8)
```
SethTisue, line 202: it seems unusual to me to bother configuring the fork join pool; wouldn't the default behavior be fine...?

SethTisue: (Even if you feel that it's better that way and don't want to remove it, at minimum I think it should be justified somehow in the article.)

docs/2024/puzzles/day06.md
```diff
@@ -1,8 +1,214 @@
 import Solver from "../../../../../website/src/components/Solver.js"
 
 # Day 6: Guard Gallivant
+# Day 06 - solution
```
SethTisue, line 4: please credit yourself as author, as seen in the previous days

makingthematrix (approved):
I would argue that the guard's  path is never infinite - in the sense that there is only a finite number of possible coordinates - and so it does not require a `LazyList` to be implemented, but I like the idea :)

docs/2024/puzzles/day06.md
samuelchassot: Yeah, you're right, makes sense. Let me remove it!
(I was experimenting with larger number of threads while writing the solution, and it is a leftover)

samuelchassot:
> I would argue that the guard's path is never infinite - in the sense that there is only a finite number of possible coordinates - and so it does not require a `LazyList` to be implemented, but I like the idea :)

Yes, if you 1) limit to the inside of the lab, which the first method of the `Guard` class does not, and 2) if you represent the coordinates as a set, and don't care about the path itself :) The path is infinite in case of loops, even if repeating 👍

SethTisue:
Thank you!

PR #641, opened 2024-12-07
Add article for day 07

Philippus:


SethTisue:
Thank you @Philippus ! I had volunteered for day 7, but I saw your PR in time and at https://github.com/scalacenter/scala-advent-of-code/discussions/571 I moved my volunteer slot to day 10 instead. If you're going to submit any further articles, please check the schedule.

Philippus:
> Thank you @Philippus ! I had volunteered for day 7, but I saw your PR in time and at #571 I moved my volunteer slot to day 10 instead. If you plan to submit any further articles, please check the schedule first.

Ah, I'm sorry. I didn't know about the schedule.

SethTisue:
@Philippus you might enjoy:  (SPOILER)

<details>

instead of processing the numbers in the equation from left to right, I process them from right to left. this allows you to prune the search space a lot and not test all possible equations. for example if you've got: `12345: ..... 2`  you can see that 12345 isn't divisible by 2, so you don't need to consider equations ending in `* 2`, and 12345 doesn't end in 2 so you don't need to consider equations ending in `| 2`. so you can recurse only on the possibility of addition and continue with `12343: ...`
</details>

PR #666, opened 2024-12-08
Update day01.md

jportway:
added day 1 solution


jportway:
appolgies for all these individual pull requests - I probably should have merged them into a single PR

PR #673, opened 2024-12-09
Add article for day 9

dyvrl:


docs/2024/puzzles/day09.md
```diff
@@ -1,15 +1,320 @@
...
+    val zippedGroups = fileFreeGroups.zipWithIndex
+    val disk = zippedGroups.flatMap:
+      case (List(fileN, freeN), i) => 
+        // File block followed by free block
+        List.fill(fileN)(Some(i)) ::: List.fill(freeN)(None)
+      case (List(fileN), i) => 
+        // Final file block
+        List.fill(fileN)(Some(i))
+      case _ => Nil
+    return disk
```
makingthematrix, line 72: No need for `return` here, just make `zippedGroups.flatMap` from above the last expression in the method.

docs/2024/puzzles/day09.md
```diff
@@ -1,15 +1,320 @@
...
+      disk.head match
+        case None => compactRec(disk.last +: disk.tail.init, acc) // Take the last element, put it first and eliminate free block
+        case file@Some(_) => compactRec(disk.tail, acc :+ file) // Append the file block
+  compactRec(disk, Vector.empty)
+```
+
+### Part 2
+The code remains very similar to `part1`. However this time, the `Disk` structure can't consider characters individually anymore. Consecutive file blocks are indivisible, they form a single `Block`. Thus, we define a new `Block` enumeration. All `Block`s have a size, but `Free` blocks do not have any index attached whereas `File` blocks do:
+```scala
+enum Block(val size: Int):
```
makingthematrix, line 96: Just a cosmetic issue, but here you use 4 spaces for an indent while everywhere else it's 2.

docs/2024/puzzles/day09.md
```diff
@@ -1,15 +1,320 @@
...
+    val intInput = input.toList.map(_ - '0') // Convert characters to int [0, 9]
+    val fileFreeGroups = intInput.grouped(2).toVector // Last group will contain a single element
+    val zippedGroups = fileFreeGroups.zipWithIndex
+    val disk = zippedGroups.flatMap:
+      case (List(fileN, freeN), id) => 
+        Vector(Block.File(fileN, id), Block.Free(freeN))
+      case (List(fileN), id) => 
+        Vector(Block.File(fileN, id))
+      case _ => Nil
+    return disk
```
makingthematrix, line 138: Let's get rid of the `return` here as well.

makingthematrix (changes requested):

SethTisue:
Thank you!

SethTisue:
fwiw, my day 9 solution is highly similar, but I managed to do without index-based logic for the recursion in part 2, by using `span`. you might enjoy: https://github.com/SethTisue/adventofcode/blob/main/2024/src/test/scala/Day09.scala#L65

PR #688, opened 2024-12-10
Title was misplaced

rmarbeck:
Moving title above link to solution writer


rmarbeck:
I guess it is preventing left menu to display Day 9 as it should

SethTisue:
thank you!

PR #747, opened 2024-12-17
Add Philippus' solution for day 06

Philippus:


SethTisue:
better late than never 😉

PR #754, opened 2024-12-18
day 18 writeup

TheDrawingCoder-Gamer:


docs/2024/puzzles/day18.md
```diff
@@ -1,11 +1,181 @@
...
 # Day 18: RAM Run
+by [@TheDrawingCoder-Gamer](https://github.com/TheDrawingCoder-Gamer)
 
 ## Puzzle description
 
 https://adventofcode.com/2024/day/18
 
+## Solution Summary
+
+1. Parse input into a List of points
```
SethTisue, line 12:
```suggestion
1. Parse input into a `List` of points
```

docs/2024/puzzles/day18.md
```diff
@@ -1,11 +1,181 @@
...
+2. For part 1, take the first 1024 points and path find through them
```
SethTisue, line 13:
```suggestion
2. For part 1, take the first 1024 points and path-find through them
```

docs/2024/puzzles/day18.md
```diff
@@ -1,11 +1,181 @@
...
+3. For part 2, find the tipping point where pathfinding can't be done anymore
+
+## Part 1
+
+First, let's make the Vec2i class for our points:
```
SethTisue, line 18:
```suggestion
First, let's make the `Vec2i` class for our points:
```

docs/2024/puzzles/day18.md
```diff
@@ -1,11 +1,181 @@
...
+
+```scala
+def parse(str: String): (Int, Int, List[Vec2i]) =
+  val ps = str.linesIterator.map:
+    case s"$x,$y" => Vec2i(x.toInt, y.toInt)
+  .toList
+  (if ps.length == 25 then 7 else 71, if ps.length == 25 then 12 else 1024, ps)
+```
+
+For both parts, we'll need a path finding function. While I would usually use a grid, because of part 2 it's easier to pathfind using a Set.
```
SethTisue, line 39:
```suggestion
For both parts, we'll need a path finding function. While I would usually use a grid, because of part 2 it's easier to pathfind using a `Set`.
```

docs/2024/puzzles/day18.md
```diff
@@ -1,11 +1,181 @@
...
+  val (size, take, walls) = parse(str)
+  walls.take(take).toSet.search(size).get.size - 1
+```
+
+(We subtract 1 here to get the number of moves made, instead of the number of tiles.)
+
+## Part 2
+
+Part 2 is finding the point where the path is impossible to reach. A naive approach would be a linear search, and with the input size it is perfectly acceptable
+with only taking 2 seconds, but a better approach is a binary search.
```
SethTisue, line 91:
```suggestion
(only taking 2 seconds), but a faster approach is binary search.
```

docs/2024/puzzles/day18.md
```diff
@@ -1,11 +1,181 @@
...
+
+All of part 2's code is entirely within it's code segment:
```
SethTisue, line 93: not sure what this means?

docs/2024/puzzles/day18.md
```diff
@@ -1,11 +1,181 @@
...
+
+```scala
+def part2(str: String): Vec2i =
+  val (size, take, walls) = parse(str)
+  Iterator.iterate(take -> walls.size): (i0, i1) =>
+    if walls.take((i0 + i1) / 2).toSet.search(size).isEmpty then i0 -> (i0 + i1) / 2 else (i0 + i1) / 2 + 1 -> i1
```
SethTisue, line 99: suggest splitting into two lines (before `then`)

docs/2024/puzzles/day18.md
```diff
@@ -1,11 +1,181 @@
...
+  .flatMap: (i0, i1) =>
+    Option.when(i0 == i1)(walls(i0 - 1))
+  .next()
+```
+
+This code will need some explaining. The iterate function is iterating through a binary search: it picks the midpoint of its two endpoints
+and searches the path with that amount of bytes fallen. If the result is None, meaning there is no valid path, then it pins its higher end point
```
SethTisue, line 106:
```suggestion
and searches the path with that amount of bytes fallen. If the result is `None`, meaning there is no valid path, then it pins its higher end point
```

docs/2024/puzzles/day18.md
```diff
@@ -1,11 +1,181 @@
...
+to that midpoint and repeats. If it's Some, meaning there is a valid path, then it pins the lower endpoint to the midpoint + 1. Doing this means you can
```
SethTisue, line 107:
```suggestion
to that midpoint and repeats. If it's `Some`, meaning there is a valid path, then it pins the lower endpoint to the midpoint + 1. Doing this means you can
```

docs/2024/puzzles/day18.md
```diff
@@ -1,11 +1,181 @@
...
+invalidate large swathes of combinations from having to be tested. Before rewriting my code to use binary search, it took around 2 seconds. This code takes
+around 50 milliseconds, so that's a 40x speedup just from using binary search.
+
+The flatMap(...).next() just finds the first point where i0 == i1, or where the binary search has completed. Because we've been taking, and not indexing, we have
```
SethTisue, line 111:
```suggestion
The `flatMap(...).next()` just finds the first point where `i0 == i1`, or where the binary search has completed. Because we've been taking, and not indexing, we have
```

SethTisue (commented):
just nitpicks

docs/2024/puzzles/day18.md
TheDrawingCoder-Gamer: what do you not understand here? can you be more specific

SethTisue: what is a code segment?

SethTisue: You mean the part 2 code is self-contained? It doesn't call anything else?

docs/2024/puzzles/day18.md
```diff
@@ -1,11 +1,182 @@
...
+
+    val start = Vec2i(0, 0)
+    val goal = Vec2i(gridSize - 1, gridSize - 1)
+    val cameFrom = mut.HashMap[Vec2i, Vec2i]()
+
+    val dist = mut.HashMap(start -> 0d)
+
+    val q = mut.PriorityQueue(start -> 0d)(using Ordering.by[(Vec2i, Double), Double](_._2).reverse)
+
+    while q.nonEmpty do
```
merlinorg, line 63: If this is `while q.nonEmpty && q.head != goal` then you can drop the early return and end with `q.headOption` which is slightly more elegant.

merlinorg (approved):

docs/2024/puzzles/day18.md
TheDrawingCoder-Gamer: Does the new text work: "Part 2 is just a short code block" ?

SethTisue:
Thank you!

PR #781, opened 2024-12-20
Add Day 15 write-up

shardulc:


SethTisue:
I put a call for reviewers on Discord — let's merge in 24 hours unless there is unresolved review feedback at that point

docs/2024/puzzles/day15.md
```diff
@@ -2,10 +2,402 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+…and a class to represent the state. In addition to the grid, we choose to (redundantly) track the position of the robot, since there’s only one and we might want its position often and don’t want to traverse the grid to find it every time.
+
+```scala
+class Warehouse(
+    val cells: ArraySeq[ArraySeq[WarehouseCell]],
+    val robotRow: Int,
+    val robotCol: Int)
+```
+
+*NB:* We use `ArraySeq`s instead of `Array`s to guarantee that we don’t accidentally mutate them in our program while still getting constant-time random access.
```
SethTisue, line 46: `scala.collection.immutable.ArraySeq`, I assume? I'd suggest saying so, as `mutable.ArraySeq` exists also

docs/2024/puzzles/day15.md
```diff
@@ -2,10 +2,402 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+
+### Part 1
+
+We start with the parser. First, we’ll need a way to create `Direction` and `WarehouseCell` objects from their `Char` representations. We could either write a simple pattern matcher:
```
SethTisue, line 50:
```suggestion
We start with the parser. First, we’ll need a way to create `Direction` and `WarehouseCell` objects from their `Char` representations. We could either write a simple pattern match:
```

docs/2024/puzzles/day15.md
```diff
@@ -2,10 +2,402 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+As a personal challenge, I try to parse without reading the entire input into memory at once, only using the `File.getLines()` iterator and standard iterable methods.
+
+```scala
+object Day15:
+  def parseMoves(input: Iterator[String]): Seq[Direction] =
+    // extra flatMap because the moves may be split across multiple lines
+    input.flatMap(_.map(Direction.ofChar)).toSeq
+
+  def parse(inputFile: String): (Warehouse, Seq[Direction]) =
+    val file = io.Source.fromFile(inputFile)
```
SethTisue, line 86: maybe `Using` instead of `try`/`finally`? I don't have a strong opinion here, just wondering if you'd considered it

docs/2024/puzzles/day15.md
```diff
@@ -2,10 +2,402 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+          ((newRobotRow, newRobotCol), WarehouseCell.Robot)) ++
+          (if stackLength > 0
+           // stackLength + 1 because we want the cell after the stack
+           then Seq((d(robotRow, robotCol, stackLength + 1), WarehouseCell.Box))
+           else Seq(/* only the robot moves */)))
+        Warehouse(newCells, newRobotRow, newRobotCol)
+```
+
+Some nifty Scala features at play:
+* `.lift` lets us turn a potentially exception-throwing method (`ArraySeq.apply` out of bounds) into one that returns an `Option`, which we can further `flatMap` with the `Option`-valued wall/box/empty logic.
```
SethTisue, line 135: I'm not sure this is the perfect wording, as `.lift` doesn't work on any exception-throwing method, it only (iirc) works on `PartialFunction`

SethTisue (commented):
(none of this is terribly important, but just some thoughts)

docs/2024/puzzles/day15.md
```diff
@@ -2,10 +2,402 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+
+Or we could make the enum parametric, and use the parameter to define a `Map` (which we can automatically use as a `Char => WarehouseCell` function due to its `apply` method):
+
+```scala
+enum WarehouseCell(val repr: Char):
+  case Empty extends WarehouseCell('.')
+  case Wall extends WarehouseCell('#')
+  case Box extends WarehouseCell('O')
+  case LeftBox extends WarehouseCell('[')
+  case RightBox extends WarehouseCell(']')
```
twentylemon, line 70: Remove? The start of the part 2 section mentions adjusting this model to account for the big boxes.

docs/2024/puzzles/day15.md
```diff
@@ -2,10 +2,402 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+    def updated(updates: Iterable[((Int, Int), WarehouseCell)])
+        : ArraySeq[ArraySeq[WarehouseCell]] =
+      // Construct a fresh array (*not* using the one backing the ArraySeq),
+      val newCells = cells.map(_.toArray).toArray
+                                                  // mutate it,
+      updates.foreach{ case ((row, col), cell) => newCells(row)(col) = cell }
+      // and construct a new ArraySeq backed by it. Possibly unsafe if the
+      // backing array remains accessible outside of the ArraySeq, but safe
+      // here because the method only returns the ArraySeq.
+      ArraySeq.unsafeWrapArray(newCells).map(ArraySeq.unsafeWrapArray)
```
twentylemon, line 162: Minor, might be able to avoid these unsafes by using `Seq` everywhere instead of the `ArraySeq` implementation type, and using `.toSeq` instead of the `.toArray`s. `Seq` is still immutable, so it won't compromise your morals!

docs/2024/puzzles/day15.md
```diff
@@ -2,10 +2,402 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+      file.close()
+```
+
+As for the move logic, I made a flawed assumption (a premature optimization?) in `getStackLengthFromTowards` logic that the objects to move will always be a contiguous stack of cells, so that just knowing its length would be enough. We now change the code in two ways:
+* Fix the flawed assumption: instead of computing the stack length as an `Int`, compute the positions of the objects to move as a `Seq[(Int, Int)]`. (This is still inside an `Option` representing whether the robot can move at all.)
+* Adapt to the new “wide box” logic: when attempting to move a wide box up or down, recursively check the cells adjacent to both the left and right sides. As a small refinement, we implement this double recursive call in the `LeftBox` case, and simply delegate to it from the `RightBox` case; otherwise, even a straight stack of *n* boxes would trigger *2^n* recursive calls. We may still have some duplicated positions in the result (e.g. when there are two boxes pushing on the same box) but it doesn’t matter too much.
+
+This method, now called `positionsToMove`, preserves the part 1 computation and slightly simplifies the `move` code too.
+
+```
```
twentylemon, line 236:
```suggestion
```scala
```

docs/2024/puzzles/day15.md
```diff
@@ -2,10 +2,402 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+        val positionsAndRobot = ((robotRow, robotCol) +: positions)
+        val newCells = cells
+          .updated(positionsAndRobot.map((_, WarehouseCell.Empty)))
+          .updated(positionsAndRobot
+            .map((row, col) => (d(row, col), cells(row)(col))))
+        val (newRobotRow, newRobotCol) = d(robotRow, robotCol)
+        Warehouse(newCells, newRobotRow, newRobotCol)
+```
+
+Another nifty Scala feature: automatic tuple destructuring with the `(_ ,_)` syntax and disjunctions with `|` inside a pattern matching expression.
```
twentylemon, line 270: TIL! I always crammed these into `if`s 🙈

twentylemon (approved):
Hello, I am a random person from said discord. 🫡 

LGTM, clean solution! 🚀

docs/2024/puzzles/day15.md
shardulc: Is it more idiomatic? I'd have to look up how to use it :sweat_smile: but I could

docs/2024/puzzles/day15.md
shardulc: Ah but I know the size ahead of time and I know I will be doing a lot of random accesses. `Seq` doesn’t guarantee anything about that, right? (I’ll add this to my reasoning note)

SethTisue (approved):

docs/2024/puzzles/day15.md
twentylemon: Yeah, but the typical implementation of Seq is the Vector, which offers basically O(1) everything. See [docs](https://docs.scala-lang.org/overviews/collections-2.13/performance-characteristics.html).

shardulc: My personal approach is to not rely on what the typical/default implementation is, but rather to only assume what the docs guarantee, and `Seq` and `IndexedSeq` would not quite guarantee enough. I would also have thought that `Vector`s provide O(1) random access but the [Scala 3 docs say](https://www.scala-lang.org/api/3.3.1/scala/collection/immutable/Vector.html#) it's O(log n).

SethTisue: (The [collections guide](https://docs.scala-lang.org/overviews/collections-2.13/performance-characteristics.html) calls that "effectively constant" since the base of the log is quite high, so that it can never take more than a small number of steps.)

PR #791, opened 2024-12-22
Add write-up for day 21

mbovel:


docs/2024/puzzles/day21.md
```diff
@@ -2,10 +2,243 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+  '7' -> Pos(0, 0), '8' -> Pos(1, 0), '9' -> Pos(2, 0),
+  '4' -> Pos(0, 1), '5' -> Pos(1, 1), '6' -> Pos(2, 1),
+  '1' -> Pos(0, 2), '2' -> Pos(1, 2), '3' -> Pos(2, 2),
+                    '0' -> Pos(1, 3), 'A' -> Pos(2, 3),
+)
+val numericalKeypadPositions = numericalKeypad.values.toSet
+
+val directionalKeypad = Map(
+                    '^' -> Pos(1, 0), 'A' -> Pos(2, 0),
+  '<' -> Pos(0, 1), 'v' -> Pos(1, 1), '>' -> Pos(2, 1),
```
SethTisue, line 32: I am so here for this artisanal formatting

docs/2024/puzzles/day21.md
```diff
@@ -2,10 +2,243 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+  def projX = Pos(x, 0)
+  def projY = Pos(0, y)
+
+val numericalKeypad = Map(
+  '7' -> Pos(0, 0), '8' -> Pos(1, 0), '9' -> Pos(2, 0),
+  '4' -> Pos(0, 1), '5' -> Pos(1, 1), '6' -> Pos(2, 1),
+  '1' -> Pos(0, 2), '2' -> Pos(1, 2), '3' -> Pos(2, 2),
+                    '0' -> Pos(1, 3), 'A' -> Pos(2, 3),
+)
+val numericalKeypadPositions = numericalKeypad.values.toSet
```
SethTisue, line 28: I'm really nitpicking here, but it ought to be "numeric" throughout rather than "numerical"

docs/2024/puzzles/day21.md
```diff
@@ -2,10 +2,243 @@ import Solver from "../../../../../website/src/components/Solver.js"
...
+
+def part2(input: String): Long =
+  input
+    .linesIterator
+    .filter(_.nonEmpty)
+    .map(line => minPathCost(line, 0, 25) * line.init.toLong)
+    .sum
+```
+
+On my example puzzle input, this solution runs in about 55 ms on the JVM and 3 ms on [Scala Native](https://scala-native.org/en/stable/).
```
SethTisue, line 208: Is most of that JVM startup time and/or HotSpot warmup time? If so, I'd suggest saying so, otherwise readers might get a mistaken impression.

And if not, I'm curious why Native would be so much faster.

In some contexts I would insist that any benchmark data be JMH-based (which is pretty easy with scala-cli's JMH support, actually) rather then going the `currentTimeMillis` route, but I don't want to make more work for you. But if it stays as is, I think it could use some additional wording about how seriously the numbers are to be taken.

merlinorg (approved):

SethTisue:
let's merge on the 23rd if there isn't any further review feedback by then

docs/2024/puzzles/day21.md
mbovel: Performance section added ✅

Indeed, the main difference is due to the lack of warm-up. However, I believe cold runtimes are just as important as hot runtimes, since not every application can afford 50k iterations before real work begins. That’s especially relevant here, where the script in a one-off. I do agree, however, that reporting both is better for completeness.

mbovel:
Thanks for the reviews! Comments addressed.

docs/2024/puzzles/day21.md
mbovel: Wait, my JMH benchmarks make no sense, I forgot to reset the cache between iterations 😭 I'll fix that tomorrow.

SethTisue: nerdsnipe succeeded 😈

nice improvement!

PR #822, opened 2024-12-30
day 14 article

TheDrawingCoder-Gamer:
No one wrote an article for day 14 (which I kind of understand, it sucked). 

I decided to write a new one that SHOULD be followable from someone who doesn't know what the final result will look like.

I want reviewers to think like someone who hasn't seen the Christmas tree output before because it's really important that you don't need to know what it looks like to get to my result.

I also removed my writeup link here.


docs/2024/puzzles/day14.md
```diff
@@ -1,11 +1,312 @@
...
+101 * 103 unique positions, or 10,403, and because all robots are moved at the same time there will only ever be 10,403 unique states. This lets us
+fail fast while writing our code in case we messed something up.
+
+
+This is arbitrary and only really possible by print debugging your code, but here's my final code:
+
+```scala
+extension (robots: List[Robot])
+  def findEasterEgg: Int =
+    (0 to 10403).find: i =>
```
twentylemon, line 119: maybe more instructive to leave 10403 as `101 * 103`, or even `size.x * size.y`

docs/2024/puzzles/day14.md
```diff
@@ -1,11 +1,312 @@
...
+def part1(input: String): Int = parse(input).stepN(100).safety
+```
+
+## Part 2
+
+Part 2 wants us to find an image which is really hard. Thankfully, there is one thing I know about Christmas trees: They have
+a lot of lines in a row and are an organized shape.
+
+The christmas trees I know are really tall, but only in a few columns. So let's test the vertical columns for a few lines that are really long.
+They also have a lot of shorter horizontal lines, so let's also check the rows for a lot of shorter lines.
```
twentylemon, line 106: Another assumption here is that the tree itself is solid; made up of many robots adjacent to each other. It's not like, an outline of a tree shape.

twentylemon (approved):
Looks good -- I'm surprised no one took this day, it was one of my favourites this year.

docs/2024/puzzles/day14.md
TheDrawingCoder-Gamer: I don't know how it would be possible to detect that if that was the case. Probably something along the lines of a flood fill but that would cost too much

twentylemon: Agreed, yeah. I meant moreso it's worth being explicit and listing it as an assumption here.

TheDrawingCoder-Gamer: Added section beforehand specifying assumption being made

SethTisue:
thank you!

TheDrawingCoder-Gamer:
Hello, you should make sure that the other person who reserved this one doesn't still want to write it. I forgot about this for a week

PR #835, opened 2024-12-30
Update path to solution

rolandtritsch:


adpi2:
Next time, I would appreciate a single PR to update all files at once. Thanks for your understanding :)

rolandtritsch:
Yes. Absolutely. Was not sure, if that would be ok or if I break the
process by doing so. - RT

PR #932, opened 2025-12-12
Fix JS widget for day 8

mbovel:
- Sync solutions submodule
- Add `day08-part1` and `day08-part2` to the solver


mbovel:
Thanks Seth! I confirm that JS solvers are now working on the latest published version.

PR #935, opened 2025-12-13
add solution to day 10

tonivade:


SethTisue:
need to target `website` branch rather than `main`

AlexITC:
@SethTisue if there isn't a reason to keep `main` as the default branch, I'd suggest to set the default branch to `website`

SethTisue:
>  if there isn't a reason to keep main as the default branch, I'd suggest to set the default branch to website

iirc it's @bishabosha and/or @adpi2 who set things up, not sure if they have any objection to changing it?

bishabosha:
I agree that the website branch is probably more important - i dont have the rights to change but no objection

SethTisue:
I don't have the rights to change it either — @sjrd?

